PR #23, opened 2017-08-28
Send xhrio request after binding event handler for response.

gardnervickers:


gardnervickers:
I'm still seeing this problem without wrapping the event handlers in `(do ...)`. I suspect there's something going on here with the emitted JS code.

PR #41, opened 2017-10-25
Small fixes to the website

fatihict:


awkay:
Thanks. I did these by hand. The image paths needed to be relative to make dev mode on them work as well.

PR #61, opened 2017-11-20
Fix register-tool

wilkerlucio:


awkay:
Yeah, I suspected there was no way they could be right without some kind of tweaking :)

PR #70, opened 2017-11-23
Fix iFrame implementation on Firefox

### LINKED ISSUE #69: Fix iFrame rendering on Firefox
The current iframe implementation is broken on Firefox:

screenshot 2017-11-23 00 46 12

The reason is that in Firefox the `iframe` node is not immediately ready to accept children appended right away. Instead, we have to wait for an `onLoad` event to be triggered before trying to add any nodes to it.

wilkerlucio:
Closes #69

### DIFF
--- a/src/main/fulcro/ui/elements.cljc
+++ b/src/main/fulcro/ui/elements.cljc
@@ -1,6 +1,7 @@
 (ns fulcro.ui.elements
   (:require [om.next :as om :refer [defui]]
-            [om.dom :as dom]))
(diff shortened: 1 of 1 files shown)


wilkerlucio:
not ready

awkay:
Did you run the existing devcards and make sure the twitter bootstrap stuff didn't break with this change?

awkay:
Nevermind...I'll try it out. If there's a problem I'll ping you :)

PR #76, opened 2017-11-27
The example go function uses the path argument

kgxsz:
It appears that the example go function in the getting started doc never made use of the path argument.

### DIFF
--- a/GettingStarted.adoc
+++ b/GettingStarted.adoc
@@ -1342,14 +1342,14 @@ In `src/dev/user.clj`, we'll add the following for development use:
 (defn go
   "Start the server. Optionally supply a path to your desired config. Relative paths will scan classpath. Absolute
   paths will come from the filesystem. The default is config/dev.edn."
-  ([] (go :dev))
+  ([] (go "config/dev.edn))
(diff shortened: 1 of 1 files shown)


GettingStarted.adoc
```diff
@@ -1342,14 +1342,14 @@ In `src/dev/user.clj`, we'll add the following for development use:
 (defn go
   "Start the server. Optionally supply a path to your desired config. Relative paths will scan classpath. Absolute
   paths will come from the filesystem. The default is config/dev.edn."
-  ([] (go :dev))
+  ([] (go "config/dev.edn))
```
awkay: missing a quote?

awkay:
Yeah, this was copy-pasted code from something else that had a map of paths, so the kw was a lookup...it needs to be cleaned up and simplified...there's no need for getting started to even have it be configurable.

GettingStarted.adoc
kgxsz: Yep, a PR after a few drinks will do that to you 😅

awkay:
I manually patched this

PR #88, opened 2017-12-05
drop use of cljs.anaylzer

thheller:
It is not DCE'd and was only used to get the fully qualified name of a `defui`. This can be done without it.

Saves at least 80kb optimized JS and a lot of compile time during dev since the entire compiler infrastructure no longer needs to be compiled in.


wilkerlucio:
Nice! One thing, you should ask to merge on the `develop` instead of `master`. Great stuff!

thheller:
Done.

awkay:
Thanks!

PR #91, opened 2017-12-07
Fix typos in README and CONTRIBUTING

andrestylianos:
Fixing a couple of typos I found while looking around.

Kudos for the great work, the framework is looking really nice and it's awesome to see the care that went into writing all the documentation.


awkay:
Thanks for the feedback and contribution!

PR #97, opened 2017-12-12
Mutation merge target

### LINKED ISSUE #66: 2.0: Consider supporting `returning` on queries without an ident
Merging into the root node of the graph is possible, so perhaps we should allow queries that have no top-level ident to be the target of `m/returning`. 

Pros: you could define an arbitrary query that could update multiple things in the root, which are easy to query via link queries.
Cons: You can easily pollute/overwrite things in the root node, where accidental collisions are too easy

wilkerlucio:
This implements most of the things we discussed. Now the meta is fully preserved when round-tripping from `query->ast` / `ast->query`.  I added that `query->ast1` helper, was handy :)

Added the `with-target` as we discussed, I changed `returning` so the order of calls between `returning` and `with-target` doesn't matter, works both ways.

A basic `:replace` with the target is working, but not the other special target values, I got tired and left that for later, so you can help me with it :)

* [x] Finish the target support so it is consistent
* [x] Write a new demo that shows it working
* [x] Run all of the demos, todoMVC, template, and devguide against it to make sure there are no regressions. This touches some sensitive stuff.

Closes #66

### DIFF
--- a/src/demos/cards/card_ui.cljs
+++ b/src/demos/cards/card_ui.cljs
@@ -25,6 +25,7 @@
     cards.server-query-security
     cards.server-return-values-as-data-driven-mutation-joins
     cards.server-return-values-manually-merging
+    cards.server-return-values-targeting
     [fulcro.client.logging :as log]))
 
 (log/set-level :debug)
--- a/src/demos/cards/server.clj
+++ b/src/demos/cards/server.clj
@@ -16,6 +16,7 @@
             cards.server_query_security
             cards.server_return_values_as_data_driven_mutation_joins
             cards.server_return_values_manually_merging
+            cards.server-return-values-targeting
             cards.loading-data-targeting-entities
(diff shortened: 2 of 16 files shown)


awkay:
Great, I'd say we should do the following still:

- [ ] Run all of the demos, todoMVC, template, and devguide against it to make sure there are no regressions. This touches some sensitive stuff.
- [ ] Write a new demo that shows it working
- [ ] Finish the target support so it is consistent

I don't want to put this on develop and possibly block other progress until we're at least sure of (1).

wilkerlucio:
@awkay Just finished adding the test, I walked all the demos, they all seem to keep working, if you can double check would be great, I think we have it ready now 😄

awkay:
You mind squashing this all down to a single commit? I could do it, but then it would say it was done by me.

wilkerlucio:
@awkay squashed

PR #106, opened 2017-12-21
Support initialize on load

awkay:
This branch is not ready to merge until it is tested in the wild a bit. We may want to change the implementation.

I'll add some comments in the delta around this.

This addresses issue 83.

### DIFF
--- a/src/main/fulcro/client/data_fetch.cljc
+++ b/src/main/fulcro/client/data_fetch.cljc
@@ -55,8 +55,9 @@
 (defn load-params*
   "Internal function to validate and process the parameters of `load` and `load-action`."
   [server-property-or-ident SubqueryClass {:keys [target params marker refresh parallel post-mutation post-mutation-params
-                                                  fallback remote without]
-                                           :or   {remote :remote marker true parallel false refresh [] without #{}}}]
+                                                  fallback remote without initialize]
+                                           :or   {remote :remote marker true parallel false refresh [] without #{}
+                                                  initialize false}}]
   {:pre [(or (nil? target) (vector? target))
          (or (nil? marker) (bool? marker) (keyword? marker))
          (or (nil? post-mutation) (symbol? post-mutation))
@@ -79,6 +80,7 @@
      :without              without
      :post-mutation        post-mutation
      :post-mutation-params post-mutation-params
+     :initialize           initialize
      :refresh              (computed-refresh refresh server-property-or-ident target)
      :marker               marker
      :parallel             parallel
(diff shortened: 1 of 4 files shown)


src/main/fulcro/client/data_fetch.cljc
```diff
@@ -79,6 +80,8 @@
      :without              without
      :post-mutation        post-mutation
      :post-mutation-params post-mutation-params
+     :initialize           (when (and initialize SubqueryClass server-property-or-ident (prim/has-initial-app-state? SubqueryClass))
+                             {server-property-or-ident (prim/get-initial-state SubqueryClass (if (map? initialize) initialize {}))})
```
awkay: @wilkerlucio This is where we could decide to make the map just literal, instead of parameters to `get-initial-state`. Thoughts?

wilkerlucio: Sounds right, so we can have the custom initialization if needed. 👍

src/test/fulcro/client/data_fetch_spec.cljc
```diff
@@ -436,6 +453,26 @@
            "Updates the global loading marker"
            @globally-marked => true)))))
 
+#?(:cljs
+   (specification "Query Response with :initialize (load ... {:initialize true})" :focused
```
wilkerlucio: I think we should also exercise the case where the response doesn't come from the server here, and instead, we see the child initial-state coming from `initial-state`.

src/main/fulcro/client/data_fetch.cljc
wilkerlucio: Hmm, now I thought again, maybe is a good idea to use the map literally instead of throwing on `get-initial-state`. The reason is the later is more flexible, if you want to re-use parts of the `initial-state` you can call it yourself. On the current situation, the user has no change to override the thing if he wants to, leading to the initial problem back again, what you think?.

src/test/fulcro/client/data_fetch_spec.cljc
awkay: Agreed. i didn't want to write a lot of tests until we had decided on the API fully. Just enough to show it working.

src/main/fulcro/client/data_fetch.cljc
awkay: That was my thought. The way it is...it's less to type at the `load` call site, but it is inherently less flexible, and since it might be a complete sub-graph might be somewhat difficult to reason about. I can go either way, but the literal map seems better than using it as params to `get-initial-state`

PR #111, opened 2017-12-25
Merge new book

awkay:


DevelopersGuide.adoc
```diff
@@ -0,0 +1,11399 @@
...
+
+When you load something you will use a query from something on your UI (it is rare to load something you don't want to
+show). Since those components (should) have a query and ident, the result of a load can be sent from the server as a
+tree, and the client can auto-normalize that tree just like it did for our initial state!
+
+==== Loading something into the DB root
+
+This case is less common, but it is a simple starting point. It is typically used to obtain something that you'd want
+to access globally (e.g. the user info about the current session). Let's assume that our Person component represents
+the same kind of data as the "logged in" user. Let's write a load that can ask the server for the "current use
```
fatihict: If you format your html the output also outputs "\n" which I found weird.

fatihict: This should be mentioned in the defmutation section. This is a really nice change and shouldn't be hidden in the demos section.

fatihict: ??

fatihict: This should be a subparagraph of `27.4. Declarative Refresh`

fatihict: Some parts are already explained in the chapter about i18n

fatihict: Which general instructions ??

fatihict: `defquery-entity` macro?

fatihict: Most of this is already explained

fatihict: ?? Demo not visible

fatihict: ?? Demo not visible

fatihict: Most of this is already explained

fatihict: Most of this is already explained

fatihict: Explained before

fatihict: Explained before

fatihict: ?? Demo not visible

fatihict: Already explained

fatihict: Already explained

fatihict: Demo not loading

fatihict: Demos not loading

fatihict: Same not 15 lines above in section `27.20. Mutation Return Values`

fatihict: Also, demos broken.

fatihict: `If that isn’t try` ??

fatihict:
4
=> Starts about vdom etc.. How is the book supposed to be read? This chapter should explain vdom otherwise like in the getting started.

4.1.1
=> Any time there are adjacent elements of the same type in the DOM, they should each have a unique :key attribute. If the elements are in an array, React requires it.
Is this still required?

=> You can convert cljs data structures (recursively) to js types with (clj->js {:x {y 1}})
Should mention this is done at runtime


4.4.1. Derived Values
=> Same example as in chapter 3

General Guidelines for Derived Values
-> formatting

4.5. Controlled Inputs
=> First mention of ui-routers.

4.6. React Lifecycle Examples
=> First mention of Google Closure

Reconciler is also mentioned but not explained

4.9 => first mention of om.next

5.1 => first mention of EDN

5.2 => again about to-one to-many relation and normalization with get-query

5.3 => lamba form is required for unions, but why?
=> examples don't follow the best practise of namespacing

5.3.2 => forms support first use, reference this to a chapter. Not sure if this example can be followed without knowing form-support first.

5.4 => the example is not a single argument function call
=> Syntax quoting was explained better in the getting started section

5.6 => You still get the ident in the props, but now spouse would resolve to a map of real properties.
=> spouse?? => A bit later also a reference to person's spouse

5.7 => Didn't understand the warning text

6.1.2 intro text is duplicate

7.1 It examines the value of :people and finds it to be a vector of maps. This indicates a to-one relationship. => to-many?
=> reference defsc here

7.2 InitialAppState is referenced while that is the protocol name not used anymore since defsc

8 intro text duplicated in 7/6 etc..

8.2 again about syntax quoting
=> assumes defmethod instead of defmutation

8.3 again about cursive integration and syntax quoting

8.4.1 uses defui, expect defsc

but the internals now two things to do => missing word "have"

8.4.2 missing " in docstring causes syntax highlighting to be messed up

8.5 already described in an earlier section

8.5.1 also duplicate

8.6 (as recommended in the templates) ?? what?

8.6.5 again about idents for normalization

in chapter 8 reconciler is mentioned a few times, but the reader wouldnt know what this is/does

8.7.6 Counter application was not used earlier
=> In the example ^:intern is being used and functions suffixed with a *. This is never explained.

9 Deep Merge references this
=> Rest API was explained earlier with a great example

9.1 again about df without explaining

9.1.2 first explanation of reconciler

For example The user => formatting

that even => event

React Lifecycle, methods => weird formatting

9.2 Easy server was already explained

9.3 data-fetch only explained now.
=> link to merge!
=> already explained this

9.3.1 also explained already
=> aswell as 9.3.2

(where a to-one edge is an ident, and a to-many edge is a vector of idents) => this really shouldn't be explained anymore

NOTE: Technically data is always loaded into the root, then relocated. So, be careful not to name your top-level edge after something that already exists there! => this should stand out more
=> Also maybe a warning

Covered in Incremental Loading => reference

9.3.6 why alpha? Looks good.

(fn [ring-response] ring-response')) => the '

9.5: duplicate
9.5.1 duplicate
9.5.2 mostly duplicate
9.5.5 df/fallback not explained
General Operations section => reference this

9.5.7 not sure if this is duplicate, maybe?

9.6.2 this references pessimistic mutations which was handler literally 1 paragraphs before..

9.6.3 props is not defined in the example
=> TODO => in the section

9.7 => correct reference
=> example redundant

9.7.1 => om/get-query also in 9.7.2

9.7.3 demo todo

9.8.1 talks about the :remote key while defmutation doesn't have that as default (in the NOTE: section)

9.8.2 these arguments are already listed expect for :marker

9.10.3 initial state of blocking component is not used by calling get-initial-state or something

9.10.4 about error recovery is the same as prior chapter about errors

9.10.6 is duplicated

10 You’re already seen a little on how to build => what is this sentence

10.1 dev guide talks about => not true

10.2 duplicate

10.2.2 same as 9.2.2

10.2.8 reference bidi

- If the API handler returns nil, => weird formatting

See the docstrings on new-config or the documentation on the easy server for more details. => reference

11.1 reference these

The example below card => ??

11.3 An environment containing => formatting bullet points

meaning of life the universe => formatting (comma)

(the following is live code): => is it?

11.3.3. A Non-trivial, recursive example => friends should a vector instead of a set

11.4 dynamic queries => reference the chapter

11.7 already mentioned

place the screen from (4) front-and-centerce the screen from (4) front-and-center => (4) ??

"foremost" => must me escaped

See the fulcro-template on github. It supports HTML5 routing with a demo tree. => reference

<→ => weird formatting.. <-> becomes that probably

13.1: important note can be formatted differently

 --- outputs --→ => again weird formatting

The rendering of the form is pretty much up to you! Thus, your forms can be as pretty (or ugly) as you care to make them. No worrying about figuring out how we render them, and then trying to make that look good. => why is this bold? Maybe use important or note?

The example shown above is doing this on application start => there is no example above

13.8: If you look carefully at PersonForm you’ll see the button to trigger adding a phone number, where we’re using (prim/tempid) to generate a temporary ID for the new phone number. => ????

13.8.3: The code also shows how you would compose the checks. The dirty? function combines the results of the nested forms together with the top form. You could do the same for validations. => which code??

13.8.3. save button??

13.10: Any you want to define : This is a map. Put whatever else you want in this map to help with rendering (e.g. placeholder text, class names, style, etc). ???

13.12: example is confusing.. Not displayed that the locally created form can go to the server

see also on-form-change => reference

iff => abbreviation all caps

\"global\" => weird formatting

see the source for defmutation validate-form => reference this

see later sections => reference

see the earlier function => reference

13.4.2 (where f is the forms namespace) => 3/4th time this is repeated
=> use defmutation instead? (whole section)

The entity of the form has a REAL ID => maybe write this as not a tempid

NOTES: => formatting

See New Relations and Removed Relations below. => reference

NOTES: * New entity properties include only the columns declared in the form support. Remember that you can declare fields without rendering them. * New entity entries do not include references! Any reference changes are always expressed with linkage change entries, as described below.
=> Formatting

who-form reasoning => ??

13.15.1 weird formatting

14.2.2 normalized-db twice in the example?

14.4 I have no idea what Nashorn is
You have two options is said, but only 1 is mentioned

Some notes: => Again a different format

There is a nashorn branch on the github fulcro-template => reference

The polyfill is just => polyfill not explained

15.1.4 reference to load markers

15.1.4 last example what is file?

16 Sente ??

16.1.1 example core/make-fulcro-server core/ ?

16.1.2 wn/ ?

17.2 namespace? instead of ns

See the formatJS documentation for further details. => reference

17.7 why is "es" assoced as a string instead of a keyword?

17.8.1 the first example is not being shown

17.9.1 sending off the locale-specific .po files => What do you mean with "sending off"?

17.10 translations package?

17.13 already mentioned in 17.10

18 reference some of dev cards to this chapter

to make cards that should UI components in all of their valid states => weird sentence

18.2 reference fulcro-css

19.1 2 two things om/tempid instead of fulcro/tempid and shouldn't this be unquoted? ~

19.2 reference See Fulcro TodoMVC for an example. also in 19.3

to have to have => reads weird

See Pre-loaded routes below. => reference

20.1.1 application-loaded!

20.2 uses defui
FIXME

21. maybe this should be introduced before 20 because it is referenced there and also on various other chapters

expected set-query! as a demo at the very least in this chapter


support viewer => maybe a section of when/what is serializable?

21.1 both examples say thing + a

22.1 + 2 todo

23.1 reference dynamic queries and routing chapters

See the templates for examples => reference

26. also outputs "\n"

we’re not handling kb events => had to think about this abbreviation

27.1.2 weird formatting defn

Closure => Google Closure

27.4 Really should also be mentioned in the defmutation section
(dc/mkdn-pprint-source Right) => ???

27.5 maybe put this as 27.4.1 or .2

27.6 some parts duplicate of the i18n section
=> also link to that section

27.7 See the general instructions for the demo => ???

27.7 defquery-root?

27.7 duplicate

27.8 " (dc/mkdn-pprint-source load-random-person) " ?? Mostly duplicate

 " (dc/mkdn-pprint-source Root) " => ??

lazy-load-tab => parens formatting

27.9 mostly duplicate

27.10 check if this is referenced

27.13 reference this on some places

27.14 computed has been described before

27.15 TODO

27.16.1 already handled in server hooks

27.17  " (dc/mkdn-pprint-source Item) " and duplicate

27.18 ")) ?? => and mostly duplicate

27.19.1 again a demo not loading

27.19.2 demos broken and reference see load documentation

27.19.3 random " at the end

27.20 reference mutation join
What is this about?????

27.20.1 same note as 27.20

demos broken

See Flyway documentation for naming conventions reference flyway and hikari

27.21 reference fulcro sql to this

DevelopersGuide.adoc
```diff
@@ -0,0 +1,11425 @@
...
+
+When you load something you will use a query from something on your UI (it is rare to load something you don't want to
+show). Since those components (should) have a query and ident, the result of a load can be sent from the server as a
+tree, and the client can auto-normalize that tree just like it did for our initial state!
+
+==== Loading something into the DB root
+
+This case is less common, but it is a simple starting point. It is typically used to obtain something that you'd want
+to access globally (e.g. the user info about the current session). Let's assume that our Person component represents
+the same kind of data as the "logged in" user. Let's write a
```
fatihict: This representation does not show the data structure above it. Joe should point to Heather and vice versa. Julie shouldn't have a spouse. Julie instead of Heather should be a child of Joe.

awkay: So, github isn't showing your comments in the right place. If you have any more, copy a sentence from the actual doc and include it for reference so I can easily find where you're looking...you could also just copy the section heading wih like "==== Idents"...just something distinct I can find.

This is fixed and I did another forced push on book branch to get some other patches in from develop.

awkay:
I'm calling this good. Merged into develop. We can open issues or new PRs to fix additional stuff.

PR #114, opened 2018-01-02
Fix a bunch of typos in the book

fatihict:


DevelopersGuide.adoc
```diff
@@ -6040,7 +6039,7 @@ Fortunately, Fulcro actually makes handling this case relatively easy as well. H
 
 1. Write a custom networking implementation for the client that detects the *kind* of error, and retries recoverable ones
 until they succeed. Possibly with exponential backoff. (If an infinite loop happens, the user will eventually hit reload.)
-2. Make your server mutations idempotent, so that a client can safely re-apply one that fails.
+2. Make your server mutations are idempotent, so that a client can safely re-apply the one that fails.
```
awkay: This one was correct as it was.

awkay: Mmmm....the comma could go away

DevelopersGuide.adoc
```diff
@@ -8057,7 +8056,7 @@ initial-form (f/validate-fields (f/build-form MyForm my-entity-props))
 ```
 
 The functions `valid?` and `invalid?` honor the tri-state nature of forms (e.g. `invalid?` returns
-true iff at least one field is `:invalid`, and `valid?` returns true iff all fields are
+true IFF at least one field is `:invalid`, and `valid?` returns true IFF all fields are
```
awkay: IFF isn't an acronym, but it also isn't good form probably...should be "if and only if"

wilkerlucio: maybe better to expand this one, I would have no idea what `iff` means.

DevelopersGuide.adoc
```diff
@@ -10705,13 +10702,13 @@ the database! The `computed` function sends the parent-generated data to the com
...
 
-So, in our list example let's talk about removing and item.
+So, in our list example let's talk about removing an item.
 Removing an item from a list requires at least one change to the database, possibly two:
 
 * You'll always want to remove the ident from the list itself. This will cause the item to disappear from the screen. This
 is sufficient if the item is being removed from the screen, but not from the db. Remember that you're working
 on a normalized database, so you have to remove an ident, not the item in this case.
-* If the item is provably being deleted from storage altogether, then you'll want to remove it from the database
+* If the item is probably being deleted from storage altogether, then you'll want to remove it from the database
```
awkay: "provably" is what I meant, not "probably"

awkay: http://www.dictionary.com/browse/provably

awkay:
Great. Yeah, if you could fix according to the comments in a new commit that would be great.

PR #117, opened 2018-01-04
Escape apostrophe in app-data

thosmos:
Fixes a problem where an apostrophe in app-state closes SSR window.INITIAL_APP_STATE prematurely


awkay:
Hi, PRs should be done against (and targeted to) develop

awkay:
What do you think of using pr-str and read-string for this instead of transit? Transit is more compact, but pr-str would just work and return a quoted thing.

thosmos:
I used pr-str with the one production app that I did SSR on.  Either way is probably fine.  I'm not sure what the risks are with using user-generated data either way though.  I also like using transit for its compactness, and it only requires this one workaround due to the single quoting.

awkay:
OK, we'll just go with the patch for now :)

PR #121, opened 2018-01-09
Remove externs reference for which the file does not exist

mitchelkuijpers:
Removed externs reference because this breaks compilation with shadow-cljs which
will complain about the missing file in the jar.


awkay:
Sounds right to me.

awkay:
I'm releasing this in 2.1.0 in a few minutes...

PR #128, opened 2018-01-30
Make app mount point id consistent.

grzm:
The `mount` example directly above uses "app" as the mount point.


awkay:
Weird...how did *that* happen?

PR #137, opened 2018-02-07
Consider adding alternate parameter syntax to joins

wilkerlucio:


src/main/fulcro/client/primitives.cljc
```diff
@@ -1460,7 +1460,8 @@
                                   (seq? join) first)
                     join        (cond-> join
                                   (util/ident? join) (hash-map '[*]))
-                    [key sel] (util/join-entry join)
+                    [key sel]   (util/join-entry join)
+                    key         (if (list? key) (first key) key) ; handles {(:key {:param 42}) [:subquery]}
```
awkay: could this be written into join-entry instead? That seems the better thing to do if it really is going to be interpreted that way.

wilkerlucio: I thought of that but seemed to me a more localized change would be safer. But I'm ok changing the `join-entry` if you think there will be no other implications.

awkay: It's a hard call. It is public API. I'd make sure join-entry had a full compliment of tests (easier to generate), then make the change. If it works will all combos of join entries, then the change is safe to outside observers, fixes db->tree, and *also* fixes code that anyone might start writing in response to seeing more examples of this alternate syntax.

My primary opposition is that as soon as examples start showing up, then the entire API has to support it...so, localized change that "makes it look official" but ends up in tons of bug reports on other parts of the API is bad.

awkay: This wider consideration means this should be considered a public API change, and *all* related code should be reveiwed before accepting it. This is *not* a trivial change, even though the code to "make it work" is small for this example.

wilkerlucio: What about making some of those utils private/deprecated? It seems their use is focused on reading stuff from the query in the query format. I think we should instead recommend that people work on the AST and not on the query. What do you think?

awkay:
So, the things that need to happen:

- [ ] Determine how invasive this change will be in the overall ecosystem. What might break? Other threats?
- [ ] Determine how much of the syntax is "hidden" inside other functions, like join-key
- [ ] Figure out if we can rewrite the other utility functions in terms of the AST logic, so there is a single-source of truth

I have other priorities at the moment, so I do not consider this high priority. I really don't want examples floating in the community code of this "new syntax", because I do believe it was an accident that it works via AST code...or at least no one that wrote the rest of the code understood it was possible.

Therefore, I consider it a syntax change that *must* be evaluating in a global context, and I highly recommend *no* examples are placed in the public that leverage it until it is an accepted, documented, syntax.

Issue #138, opened 2018-02-07
tutorial throws some warnings: Each child in an array or iterator should have a unique "key" prop


yayitswei:
I'm looking at the B_UI_Exercises.cljs devcards and getting the following warnings:

![screen shot 2018-02-06 at 11 19 37 pm](https://user-images.githubusercontent.com/217094/35903494-44dfa692-0b94-11e8-8f95-03149fcd1270.png)

awkay:
Thanks for the report. If you plan to fix the warnings, a PR is appreciated. Otherwise, I have to prioritize things, and I don't leave issues open that I don't plan to do anything about soon.

yayitswei:
No worries, I understand the bandwidth issue. Thanks for making fulcro! I'm really enjoying it so far. As a newbie to the library, I don't know how to fix the warnings and was worried it would prevent completion of the exercises, but if it's not page-breaking I can live with it.

awkay:
It's a react error. Somewhere on that page is a vector, list or some other
collection of components that don't set `:key` so react can easily diff them

PR #139, opened 2018-02-09
Fix React missing :key warning

pedrorgirardi:
Hi @awkay. I saw the issue #138 and I thought I could try to fix it. I'm not sure if this is the best way to do it or even if you would bother to merge such a small thing.
Anyway. Thanks!


awkay:
Sure, that looks fine.

PR #147, opened 2018-02-20
Feature/split parser

awkay:
I renamed the parameter `:query-interpreter` and added the code to detect and split the parsing. Also removed some dead code and fixed a bug. I'm merging this and pushing it as 2.3.0-SNAPSHOT, and I'll be testing it. Let me know if you see any problems Wilker. Thanks!

### DIFF
--- a/src/main/fulcro/client.cljc
+++ b/src/main/fulcro/client.cljc
@@ -196,7 +196,7 @@
       ([env query target]
        (if (mutation-query? query)
          (mutation-parser env query target)
-         (query-parser (assoc env :target target) query)))
+         (query-parser env query)))
       ([env query] (split-parser* env query nil)))))
 
 (defn- initialize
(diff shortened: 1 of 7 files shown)


src/main/fulcro/client.cljc
```diff
@@ -184,6 +184,21 @@
...
+(defn- split-parser
+  "Generate a parser that splits reads and writes across two parsers: the supplied query parser for queries, and the built-in
+  parser for mutations."
+  [query-parser]
+  (let [mutation-parser (prim/parser {:read (constantly nil) :mutate app/write-entry-point})]
+    (fn split-parser*
+      ([env query target]
+       (if (mutation-query? query)
+         (mutation-parser env query target)
+         (query-parser (assoc env :target target) query)))
```
wilkerlucio: I thought we would only send the custom parser when there is no target, we changed the assumption or was a miss?

awkay: this is just to make sure there is a target on there...it will be nil unless I've screwed up

awkay: I guess it could just be dropped...I was thinking ppl would expect to see it, but probably not, actually

awkay:
I've removed that little bit of code. Thanks for the review. I tried this out on some basic apps and it seems to be ok, as I would expect.

PR #148, opened 2018-02-21
Check that attrs passed to dom fns are valid

ChrisBlom:
Adds assertions to the dom functions to check if the attrs argument 
is valid.
The attrs are valid if they are `nil?` or `object?` (for cljs) or `map?` (for clj)
With this you get a stacktrace that includes the line where the invalid attr was passed, 
without this assert react will give you a warning, but not enough info to locate the where the error came from.

### DIFF
--- a/src/main/fulcro/client/dom.cljc
+++ b/src/main/fulcro/client/dom.cljc
@@ -282,9 +282,14 @@
 
 #?(:clj (gen-react-dom-inline-fns))
 
+#?(:clj
+   (defn valid-opts? [opts]
+     (or (nil? opts) (map? opts))))
+
 (defn ^:private gen-react-dom-fn [tag]
   `(defn ~tag [opts# & children#]
      {:style/indent 1}
+     (assert (valid-opts? opts#) ~(str "first argument to " tag " must be either nil or a js object"))
      (.apply ~(symbol "js" "React.createElement") nil
        (cljs.core/into-array
          (cons ~(name tag) (cons opts# (cljs.core/map fulcro.util/force-children children#)))))))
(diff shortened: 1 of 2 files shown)


awkay:
Hm. So I've not played with this, but I'm aware that it is sorta possible to subclass the built-in types like Map in js now...will `object?` return true if the user has done that? I'm just trying to ensure we're not pushing a breaking change.

ChrisBlom:
I'm not sure, i'm not that familiar Javascript interop, I'll try to check under which circumstances `object?` does not return the right result.

awkay:
So, I'm considering adding support for auto-conversion from js to cljs. I want to do some benchmarking and see if it makes sense. The other option is to make a new namespace that does it, and leave the old one alone. That would ensure we don't break things, but give users a better experience when they don't need the absolute optimal performance.

I suspect that the checking overhead can be made very very low.

The alternate namespace option would make it so that this assertion is less important, since people could more easily skip over the cause of the problem to begin with.

awkay:
I'm going to defer this until we've had a chance to test the possible alternatives.

The idea change would:

- [ ] Allow users to choose to use cljs maps instead of having to use `#js`
- [ ] Be as performant as the current function when the user chooses to use `#js` (non-zero, but negligible impact)
- [ ] Not cause current programs to fail during upgrade

awkay:
I've done real benchmarking. See https://github.com/fulcrologic/fulcro/issues/153

I have decided that a cond on the dom generation functions is a great idea, and won't have any impact. I also agree (from my measurements) that it is unlikely anyone will notice a performance problem using clj maps and having them auto-convert internally.

awkay:
After benchmarking, I decided to start a new PR: https://github.com/fulcrologic/fulcro/pull/154

If anyone wants to help get the final touches on that, let me know. The wrapped inputs still need to be fixed. I'd also like wrapped inputs to support `:ref` functions, which they currently screw up somehow.

Issue #143, opened 2018-02-13
Review new websockets


awkay:
Client fulcro/websockets.cljs
- [x] Test Untested stuff
   - [x] req-params
   - [x] transit-handlers
- [x] Messy code around API response handling
- [x] global-error-callback missing will throw an exception
- [x] Anything you see
- [ ] auto-rety parameters...retry limit?

Server fulcro/websockets.clj
- [x] wrap-api isn't configurable on URI
- [x] Move URI to websockets component and make wrappers use it (no hard-coded /chsk)
- [x] Other sente concerns??? Do we want to handle more events?

Finally, are there any other extension points ppl will want?

currentoor:
https://github.com/fulcrologic/websockets2-demo/pull/1#issuecomment-366158243

I like how on the client you merge custom transit handlers into the tempid ones needed for fulcro. But allowing users to pass in the packer is the most customizable. I see two options

1. Users are allowed to pass in a packer, and fulcro comes with a helper function for creating a packer (otherwise the default packer is used).
```cljs
(defn create-packer [transit-handlers] ...)
```

2. Users pass in transit handlers on both sides, which get merged with the default tempid ones, and the the result is used to construct a packer. Of course sente the sente options map will override these if a packer is included.

Do you have a preference? 2 seems easier for the end user, but 1 seems simpler. I'd rather do 1. Or a different approach?

Also, I'm happy to code this up and update the demo also.

awkay:
Is there an objective technical advantage to packers? Our user base is more
accustomed to transit.

currentoor:
I worked on a project where they had a packer that used edn for small
messages and transit for larger ones. That’s the only technical advantage I
can think of. But that, was probably more of a pre-optimization rather than
proper use case. I doubt it was worth the complexity.

Just transit handlers on both sides is probably good enough. You want me to
do that?

awkay:
Yes, please :)

awkay:
All fixed in 2.3.0. NOTE: API changed slightly.

PR #149, opened 2018-02-22
Made ws uri configurable and added transit handlers

currentoor:
Tested here https://github.com/fulcrologic/websockets2-demo/pull/2.

For https://github.com/fulcrologic/fulcro/issues/143

- Makes the socket uri configurable, client side was already configurable
- Adds a transit handlers options, consistent with what was available in the client side component
- Other sente events did not seem worth handling

**Potentially TODO**
- global-error-callback missing will throw an exception, when does this happen?
- auto-rety parameters...retry limit?, sente does an exponential back off for retries so maybe it's not needed?
- Messy code around API response handling (seems fine to me)

### DIFF
--- a/src/main/fulcro/websockets.clj
+++ b/src/main/fulcro/websockets.clj
@@ -37,7 +37,8 @@
                              (log/error "Reply function missing on API call!")))
       (do :nothing-by-default))))
 
-(defn- is-wsrequest? [{:keys [uri]}] (= "/chsk" uri))
+(defn- is-wsrequest? [{:keys [websockets-uri]} {:keys [uri]}]
+  (= websockets-uri uri))
 
 (defrecord EasyServerAdapter [handler websockets]
   component/Lifecycle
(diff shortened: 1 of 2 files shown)


src/main/fulcro/websockets.clj
```diff
@@ -132,21 +138,21 @@
 
   If you don't supply a server adapter, it defaults to http-kit.
   "
-  ([parser]
-   (make-websockets parser nil {}))
-  ([parser http-server-adapter sente-socket-server-options]
-   (map->Websockets {:server-options (merge {:user-id-fn (fn [r] (:client-id r))} sente-socket-server-options)
-                     :server-adapter (or http-server-adapter (hk/get-sch-adapter))
-                     :parser         parser})))
+  [parser websockets-uri {:keys [http-server-adapter transit-handlers sente-options]}]
```
awkay: Seems like moving these all into a map of options might be better. Especially for optional things...so perhaps the signature should be `(fn [parser options])`?

currentoor: That was my intention too, but the `websockets-uri` is not optional. At least not right now, on both the client and server. Do you want me to make them optional on both, with default value of `"/chsk"`?

awkay:
Yes, defaulting to chsk was my intention.

PR #154, opened 2018-02-24
added cond that allows for auto-conversion of clj maps as props

awkay:
Remaining work: The DOM inputs use wrapped-dom-input, which is a crazy bit of code intended to help with lost input. Turns out it is quite useful, but I still need to add the props conversion in the correct places.

The other problem with the wrapped inputs in that they don't support the new react refs correctly. If anyone wants to look into that, let me know.


awkay:
We've decided to import Thomas Heller's fixes instead. The wrap-form-input was fixed in an alternate commit on develop.

Issue #153, opened 2018-02-24
Enhancements to dom rendering functions


awkay:
It was suggested that we add an assertion to the dom rendering functions to give an error when `#js` is forgotten. That seemed reasonable, but I've always wondered if the `#js` requirement was worth the benefit of the optimization, so I decided to try to quantify it.

I did the following experiment:

Render 10 divs containing 90 spans at 60fps, while measuring the time to render.

BEFORE code changes:
nil props: Early iterations at about 1.95ms/frame, converging towards 0.67ms/frame
3k/v  pairs in `#js` obj as props: Early iterations 2.5ms/frame, converging to 1ms/frame

AFTER code changes:
nil props: no measurable effect
JS obj 3k/v pairs: no measurable difference
CLJ obj 3k/v pairs: Early iterations 6-7ms/frame, converging to 2.07ms/frame

To ground-truth this I ran a few simple timings on map conversions, and a single k/v map seems to take about 3-4 microseconds to convert. A map with 3 entries takes about 10 microseconds (when done in units of 100 iterations). Thus, I would expect the overhead for `clj->js` in this benchmark to consume about 1ms/frame, which is exactly what I did measure.

I then measured the overall browser performance flame chart to get an idea of how long the frame was actually taking from the complete React standpoint (all DOM was needing a refresh because I was changing content).

NO (nil) props:
User code execution: 2.1ms
Overall time including browser paint: 3.5ms

CLJ props:
User code: 4.3ms
Overall time including browser paint: 6ms

So, now we can gain some perspective if we relate this to a target frame time max of 16ms as the target number to maintain full 60fps animation rates. 

Fulcro, of course, has overhead of the query engine. That is a relatively heavy task, so let's say we allocate half of this time to query engine concerns. This leaves our rendering task 8ms. 

If we assume an average update on the order of 100 DOM elements (which is generous considering shouldComponentUpdate should prevent non-essential changes), then using clj maps for properties should represent no more than about 1/8th of our usable time (assuming all nodes needed props, which again is quite generous).

User code execution in the flame graph indicates that React consumes about 1.3 to 2ms/frame for this task. 

One can conclude that most common web applications will not see a significant performance impact from using clj maps; however, if they do these measurements also show that the inclusion of the suggested `cond` in the DOM generation functions has such a small effect as to be immeasurable in the face of performance deviations.

[...]

PR #159, opened 2018-03-01
Ported fancy nested dom macros from Thomas Heller's shadow repo

currentoor:
Addresses https://github.com/fulcrologic/fulcro/issues/153

They are tested on the front end in devcards and on the back end with SSR unit
tests. Both appear to be working. Have not tried input elements.

Here is the novel part of `dom.cljc`
https://github.com/fulcrologic/fulcro/pull/159/commits/243d36542ba7d4d8194072eb6aba416befed4285#diff-e699c240383a1c0d1eb2d479c63e00d1R728.

Still a work in progress.

### DIFF
--- a/src/main/fulcro/client/alpha/dom.cljc
+++ b/src/main/fulcro/client/alpha/dom.cljc
@@ -1,0 +1,808 @@
+(ns fulcro.client.alpha.dom
+  (:refer-clojure :exclude [map meta time])
+  #?(:clj
+     (:require [clojure.string :as str]
+               [fulcro.client.impl.protocols :as p]
+               [clojure.core.reducers :as r]
+               [fulcro.checksums :as chk]))
+  #?(:clj (:import (cljs.tagged_literals JSValue))))
+
+(declare tags
+  a
+  abbr
+  address
+  area
+  article
+  aside
+  audio
+  b
+  base
+  bdi
+  bdo
+  big
+  blockquote
+  body
+  br
+  button
+  canvas
+  caption
+  cite
+  code
+  col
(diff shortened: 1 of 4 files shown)


awkay:
sweet. If there's a stopping point let me know and perhaps we can merge this into a feature branch that we can both work on...if there's work that can be split. Perhaps the id/class converter?

src/main/fulcro/client/alpha/dom.cljc
```diff
@@ -0,0 +1,808 @@
...
+;;        ~@(clojure.core/map gen-react-dom-fn tags)
+;;        ~@extra-inputs)))
+
+;; ===================================================================
+;; Server-side rendering
+
+;; https://github.com/facebook/react/blob/57ae3b/src/renderers/dom/shared/SVGDOMPropertyConfig.js
+;; https://github.com/facebook/react/blob/57ae3b/src/renderers/dom/shared/HTMLDOMPropertyConfig.js
+#?(:clj
+   (def supported-attrs
```
wilkerlucio: React itself dropped attribute validations (https://reactjs.org/blog/2017/09/26/react-v16.0.html#support-for-custom-dom-attributes) because they prevent you from adopting new features (like hurting the open principle). So I think we are better not validating it, what you guys think?

awkay: I agree

awkay:
I'm merging this to develop. It's a new namespace so  it can't hurt anything, and then we can collaborate on it. I'm really interested in this particular problem right now :)

PR #162, opened 2018-03-04
Dom API V2 CSS Shorhand

currentoor:
Still a work in progress but just wanted to push up what I had so far.

- Recursively convert map literals to js object at compile time
- Spec'd dom macro arguments
- Added CSS shorthand notation parsing

TODO
- More thorough and organized tests, there a lot of permutations possible
- Cleanup code
- Obvious compile-time/runtime errors for trying to use css shorthand with js object? Or allow that?


awkay:
I may work on writing some more tests, but for the time being you'll want to pull from upstream develop....or rather, rebase on it.

PR #185, opened 2018-04-20
Add js type hints for externs inference

levitanong:


awkay:
So, thanks for going through these. I'd prefer it if we add type hints that are accurate, and don't just suppress the warnings. There is some research to do on this. For example, on deftype data (FulcroProps), should it be type-hinted more specifically??? I have not studied js type-hinting, but it at least seems like we should add type hints like `js/React.Component` when that is what the thing is (like on shouldComponentUpdate).

awkay:
Ah, I see...I read some of the shadow-cljs manual...these won't hurt the normal clojurescript compile, right? They're just metadata...should there be an externs file for those I wonder?

levitanong:
Yeah, I've developed the whole day with this on, so it seems fine. Even when I didn't have those type hints, advanced compilation worked fine for me, so I'm assuming there's no need for externs there.

awkay:
OK, so on the i18n warnings...you didn't do anything about those, you were just reporting them?

levitanong:
I do think more specific type hints would provide some documentation benefit, but I don't know enough about all this to confidently say that there is no optimization benefit.

Yeah, I was just reporting the i18n warnings.

PR #192, opened 2018-05-16
Tiny tweak of forms chapter

OliverM:
Fixing a minor typo that tripped me up a bit.

### DIFF
--- a/DevelopersGuide.adoc
+++ b/DevelopersGuide.adoc
@@ -8514,7 +8514,7 @@ with components like:
 Normally you might throw this into your application state with something like:
 
 ```
-(prim/merge-component! reconciler Person person-tree)
+(prim/merge-component! reconciler PersonForm person-tree)
 ```
 
 which will insert and normalize the person and phone number.


awkay:
Hm. I'm wondering if more lines need editing...I see Person several more times in that section. I'd have to carefully go through it to understand what I wrote again. Why didn't you have problems with the other uses of Person?

OliverM:
I was trying to create a minimal example in my editor as I went, using what was covered form the start of the chapter until the end of section 15.2.1. I couldn't properly follow what the text was saying until that point because I couldn't see where the Person component had been defined prior to that, and didn't really get how the form support augmented the component being edited. I inferred what the Person component should be, and It wasn't until I saw the error `Attempt to add form configuration to test-ns/Person, but it does not declare any fields!` that I began to follow what was going on.

This is just a minor tweak to save some other notional person having to go down that same path, but I think you're right about the chapter in general needing more work. There are more references to the Person component, but the code examples presented come quite late in the chapter (an earlier, minimal version that only edits an already existing component with no network interaction, validation etc - just showing the plumbing - would be useful). Also the two full examples use Fulcro's Bootstrap utilities which obscure the rendering code if you're not familiar with them (also solved by a minimal early example).

I can help out with this if you think the above is a good direction to take?

awkay:
Sorry, I've been kind of busy. I'm happy for any help, but code samples in the book chapter need to be actually coded up and tested...ideally in book demos that are embedded in the book itself to verify that they continue to work (one of the things I like to do every once and a while is run through the book and make sure all of the demos are still working...in case of regression).

If you'd be interested in helping with that I'd be glad to show you around the book build and embedding technique.

OliverM:
I'd love to help out with the docs! Are there areas in particular you think need more attention than others?

awkay:
Great! The forms section you pointed out is the worst offender at the moment I think.  There is a figwheel build for book demos in the project file. You just set -Dbook on the JVM args when running script/figwheel.clj, and the use the DevelopersGuide.html file locally (which is generated by `make book` with an installed asciidoctor...which can come from homebrew). 

brew install asciidoctor

and I think you need to gem install some stuff like coderay and asciidoctor-diagram to get the book to build completely right (with code formatting and images from ascii art). 

I generally do this when working (via IntelliJ with asciidoc plugin):

in a terminal:

```
while true
do
make book
sleep 2
done
```

Then  start the figwheel build:

```
JVM_OPTS=-Dbook lein run -m clojure.main script/figwheel.clj
```

Then open `docs/DevelopersGuide.html` in a browser.

You have to reload the browser to see HTML changes, but code edits will hot reload.

There are a few steps to add a new "live demo":

1. Write it, and use the `render-example` or `ui-iframe` if you need to encapsulate external CSS. See exising examples.
2. Require it in book.main
3. In book.main, turn it into an example with my `defexample` macro
4. In the adoc of the developer's guide, you need to copy an existing example embed and edit it to match your new example.  Plenty of examples here too.  (search for `+++`, but don't use the first two or so, since they are special).

awkay:
Closing this for now.  Feel free to actually do a contribution if you get to it.

Issue #216, opened 2018-07-05
GC Support


currentoor:
We need some GC support. 

We can have a function specifically for removing an ident from a list.
`(remove-ident state-map ident path)`, where `path` is the path in app state to the list, like `prim/integrate-ident`.

We can also have a function for larger GC, `(clean-ident state-map #{table-names} ident)`. This function will remove `ident` from those tables specified. 

We should have two version of each, like `integrate-ident` and `integrate-ident!`.

Maybe we also need a way to do a deep delete and remove entities across joins. Like in this function. Perhaps that should be called `deep-clean-ident`?
https://github.com/fulcrologic/fulcro-inspect/blob/develop/src/client/fulcro/inspect/helpers.cljs#L123-L140

These examples in the book should be updated after this change.
http://book.fulcrologic.com/#_handling_mutations
http://book.fulcrologic.com/#_mutations_on_a_normalized_database
http://book.fulcrologic.com/#_triggering_the_remote_mutation_from_the_client

awkay:
Hey, small request: If you're going to open a feature issue with goals, please also include a milestone or some other kind of timeline.  I don't believe in opening issues we're not going to immediately address.  Leads to a pile of stuff that no one looks at (affectionately know in lean circles as a Jira Load :D  ).

currentoor:
Good idea. I never heard of that term, but it does seem appropriate 😄 

How about a timeline of by the end of this week, if not middle of next week at the latest?

awkay:
So we've identified a few categories of these:

1. Helper functions for targeted state manipulation when you *know* what you want to remove. Not really GC, but helpful. These would be functions like `remove-ident`, `delete-form-state`, etc.
2. Semi-automated GC...e.g.  `(recursive-delete some-ident)`, which delete the table entry at ident, and ALSO follows idents starting at that entity, and removes all children. This requires something like a white-list/black-list of table names, or a set of "isComponent" Datomic-like indicators to prevent unintentional traversal.
3. Fully-automated GC. The function scans entire app state removing things that are not referenced. This one also needs "help" in that unions may not be currently pointing at data that needs to be kept. 

A few helpers from (1) would be nice, especially as "mutation helpers".  (2) could be commonly useful...not sure about 3.

awkay:
The basic remove ident and form state are in RC4.  Others are deferred for now.

awkay:
Latest release of Fulcro has a new `normalized-state` ns that has more helpers.

PR #220, opened 2018-07-18
Added remove-ident function with specs to primitives

currentoor:
@awkay you said

> 1. Helper functions for targeted state manipulation when you know what you want to remove. Not really GC, but helpful. These would be functions like remove-ident, delete-form-state, etc.
> 2. Semi-automated GC...e.g. (recursive-delete some-ident), which delete the table entry at ident, and ALSO follows idents starting at that entity, and removes all children. This requires something like a white-list/black-list of table names, or a set of "isComponent" Datomic-like indicators to prevent unintentional traversal.
> 3. Fully-automated GC. The function scans entire app state removing things that are not referenced. This one also needs "help" in that unions may not be currently pointing at data that needs to be kept.

At the moment I only needed (1), so that's all I did. If the need for (2) or (3) arrises, I'll submit another PR.

Also, after this is merged I can update the book examples to use it. 

https://github.com/fulcrologic/fulcro/issues/216

### DIFF
--- a/src/book/book/demos/parent_child_ownership_relations.cljs
+++ b/src/book/book/demos/parent_child_ownership_relations.cljs
@@ -23,7 +23,7 @@
         (fn [s]
           (-> s
             (update :items dissoc id)
-            (update-in [:lists 1 :list/items] filter-item)))))))
+            (m/remove-ident* [:lists 1 :list/items] [:items id])))))))
 
 (defsc Item [this
              {:keys [item/id item/label] :as props}
--- a/src/book/book/demos/parent_child_ownership_relations.cljs
+++ b/src/book/book/demos/parent_child_ownership_relations.cljs
@@ -18,12 +18,11 @@
   "Mutation: Delete an item from a list"
   [{:keys [id]}]
   (action [{:keys [state]}]
-    (letfn [(filter-item [list] (filterv #(not= (second %) id) list))]
-      (swap! state
-        (fn [s]
-          (-> s
-            (update :items dissoc id)
-            (m/remove-ident* [:lists 1 :list/items] [:items id])))))))
+    (swap! state
+      (fn [s]
+        (-> s
+          (update :items dissoc id)
+          (m/remove-ident* [:lists 1 :list/items] [:items id]))))))
 
 (defsc Item [this
              {:keys [item/id item/label] :as props}
--- a/src/main/fulcro/client/primitives.cljc
+++ b/src/main/fulcro/client/primitives.cljc
@@ -3179,15 +3179,6 @@
                   (throw (ex-info "Unknown post-op to merge-state!: " {:command command :arg data-path})))))
       state actions)))
 
-(defn remove-ident
-  "Removes an ident, if it exists, from a list of idents in app state. This
-  function is safe to use within mutations."
-  [state-map ident path-to-idents]
-  {:pre [(map? state-map)]}
-  (let [new-list (fn [old-list]
-                   (vec (filter #(not= ident %) old-list)))]
-    (update-in state-map path-to-idents new-list)))
-
 (defn component-merge-query
   "Calculates the query that can be used to pull (or merge) a component with an ident
   to/from a normalized app database. Requires a tree of data that represents the instance of
@@ -3236,14 +3227,6 @@
   (assert (is-atom? state)
     "The state has to be an atom. Use 'integrate-ident' instead.")
   (apply swap! state integrate-ident ident named-parameters))
-
-(defn remove-ident!
-  "Removes an ident, if it exists, from a list of idents in app state. This
-  function is safe to use within mutations."
-  [state ident path-to-idents]
-  (assert (is-atom? state)
-          "The state has to be an atom. Use 'remove-ident' instead.")
-  (swap! state remove-ident ident path-to-idents))
 
 (defn merge-component
   "Given a state map of the application database, a component, and a tree of component-data: normalizes
--- a/src/main/fulcro/client/primitives.cljc
+++ b/src/main/fulcro/client/primitives.cljc
@@ -3141,8 +3141,8 @@
   "DEPRECATED: Use fulcro.client.mutations/integrate-ident* in your mutations instead."
   [state ident & named-parameters]
   {:pre [(map? state)]}
-  (log/warn "This function has been been deprecated and will be removed in the future."
-            "Use fulcro.client.mutations/integrate-ident* in your mutations instead.")
+  (log/warn "integrate-ident is deprecated and will be removed in the future."
+            "Please use fulcro.client.mutations/integrate-ident* in your mutations instead.")
   (apply util/__integrate-ident-impl__ state ident named-parameters))
 
 (defn component-merge-query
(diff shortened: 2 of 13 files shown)


awkay:
So, the API of remove-ident should probably include a flag to indicate if the targeted table entry was "owned" and should be cleaned up.  Probably a multi-arity version:

`(remove-ident state-map ident path true)` means remove the ident from the vector at path, and *also* `(update state-map (first ident) dissoc (second ident))`, whereas a missing 4th arg would mean the table entry stays.  Thoughts?

currentoor:
I was initially thinking this would be a low-level function only concerned with removing an ident from a path.

I usually have the logic for *removing ident from list* in one mutation and the logic for *removing the entity* in another mutation. The former is located near the UI component that needs that list of idents.

I prefer keeping them separate. Feels cleaner and more explicit. 

But if we're going to combine them, we should probably call it something else no? Or make a function called `remove-entity` which removes the entity from its table, and takes a list of paths (so calls `remove-ident` on each of those paths). What do you think?

 Also I forgot to add the form state cleanup function to this PR. I'll do that now.

wilkerlucio:
I think we should use fulcro incubator for those kinds of helpers, it's hard to figure how good they are until we put then for a trial period, and I think fulcro incubator is the right place to let these functions grow. This could be added to https://github.com/fulcrologic/fulcro-incubator/blob/develop/src/fulcro/incubator/db_helpers.cljc

currentoor:
I agree with @wilkerlucio for the higher level functions, however the lower-level ones in this PR IMHO obviously need to be in Fulcro.

The logic of `remove-ident` is already duplicated 3 times in the book. The logic in `delete-form-state*` requires you to first figure out how fulcro manages form state copies, we don't require people to figure that out when they want to create a form-state copy, so why make them figure it out when they want to remove a form-state copy.

wilkerlucio:
Makes sense, I don't have a strong opinion on this one in specific, not against having it on primitives, thanks for clarifying @currentoor !

src/main/fulcro/ui/form_state.cljc
```diff
@@ -546,6 +546,20 @@
   :args (s/cat :state map? :entity-ident util/ident? :field (s/? keyword?))
   :ret map?)
 
+(defn delete-form-state*
```
awkay: I think remove the varargs, and make it detect a vector of idents or a single ident. That way threading will still work well with it (since you cannot easily use apply in threading).

src/main/fulcro/client/primitives.cljc
```diff
@@ -3179,6 +3179,15 @@
                   (throw (ex-info "Unknown post-op to merge-state!: " {:command command :arg data-path})))))
       state actions)))
 
+(defn remove-ident
```
awkay: Move to mutations, and rename `remove-ident*`

src/main/fulcro/client/primitives.cljc
```diff
@@ -3228,6 +3237,14 @@
     "The state has to be an atom. Use 'integrate-ident' instead.")
   (apply swap! state integrate-ident ident named-parameters))
 
+(defn remove-ident!
```
awkay: Move to mutations, and make the signature:

`(remove-ident! this field ident)`

so it can be used in a component that owns a list instead of writing a transact.

awkay:
While we're at it, please copy `integrate-ident` to `mutations/integrate-ident*`, alias the old one to the new one, make the old doc string say DEPRECATED, SEE (the other one).  Also add deprecation to `integrate-ident!`and say to combine `integrate-ident*` into your mutations instead.

Add log warnings about deprecation that tell them how to port.

awkay:
Also, be sure to add to CHANGELOG, and DevelopersGuide.adoc

currentoor:
@awkay I made the edits you suggested and updated the CHANGELOG. Also, I updated the dev guide in one place. I could have updated more, but it looks you were trying to make point about how simple updating a normalized app-state is. I felt like using the mutation helper `remove-ident*` would obfuscate that point. Please let me know if you disagree.

src/book/book/demos/parent_child_ownership_relations.cljs
```diff
@@ -23,7 +23,7 @@
         (fn [s]
```
awkay: Why not delete filter-item?

src/main/fulcro/client/primitives.cljc
```diff
@@ -3138,46 +3138,12 @@
...
-                             (assert (vector? data-path) (str "Replacement path must be a vector. You passed: " data-path))
-                             (when to-many?
-                               (do
-                                 (assert (vector? vector) "Path for replacement must be a vector")
-                                 (assert (number? index) "Path for replacement must end in a vector index")
-                                 (assert (contains? vector index) (str "Target vector for replacement does not have an item at index " index))))
-                             (assoc-in state data-path ident))
-                  (throw (ex-info "Unknown post-op to merge-state!: " {:command command :arg data-path})))))
-      state actions)))
+  (log/warn "This function has been been deprecated and will be removed in the future."
```
awkay: No one is going to know what "This function" is. Well, they can infer it, but the message should say `integrate-ident is deprecated`.

src/main/fulcro/client/primitives.cljc
```diff
@@ -3211,22 +3177,13 @@
...
-  The named parameters can be specified any number of times. They are:
-
-  - append:  A vector (path) to a list in your app state where this new object's ident should be appended. Will not append
-  the ident if that ident is already in the list.
-  - prepend: A vector (path) to a list in your app state where this new object's ident should be prepended. Will not append
-  the ident if that ident is already in the list.
-  - replace: A vector (path) to a specific location in app-state where this object's ident should be placed. Can target a to-one or to-many.
-   If the target is a vector element then that element must already exist in the vector.
-  "
+  "DEPRECATED: Use fulcro.client.mutations/integrate-ident* in your mutations instead."
```
awkay: Same comment: "integrate-ident! is deprecated and will be removed in a future version. Please use..."

src/main/fulcro/ui/form_state.cljc
```diff
@@ -546,6 +546,24 @@
...
+
+        ks (mapv (fn [[t r]]
+                   {:table t :row r})
+                 entity-idents)]
+    (update state-map ::forms-by-ident
+            (fn [s]
+              (apply dissoc s ks)))))
+
+(s/fdef delete-form-state*
+        :args (s/cat :state-map map? :entity-idents (s/* util/ident?))
```
awkay: I don't think this spec is right. Should be an OR or ident or every ident

awkay:
Sorry I'm slow. Lot's going on.  Couple more comments.

currentoor:
@awkay I addressed your comments, added `sente-options` to client side websockets, and made the default sente socket type auto (which prefers websockets if available).

Also I tested the websockets changes locally in my app.

src/main/fulcro/websockets.cljs
```diff
@@ -81,10 +86,12 @@
       `state-callback` can be either a function, or an atom containing a function.
   - `global-error-callback` - A function (fn [resp] ...) that is called when returned status code from the server is not 200.
   - `auto-retry?` - A boolean (default false). If set to true any network disconnects will lead to infinite retries until
+  - `sente-options` - A map of options that is passed directly to the sente websocket channel construction (see sente docs).
   the network returns. All remote mutations should be idempotent.
```
awkay: oops. This line goes with auto-retry

src/main/fulcro/websockets.cljs
```diff
@@ -97,4 +104,5 @@
                      :transit-handlers      transit-handlers
                      :app                   (atom nil)
                      :stop                  (atom nil)
-                     :req-params            req-params})))
+                     :req-params            req-params
+                     :sente-options         sente-options})))
```
awkay: What happens if you don't pass options?  You'd be passing nil to sente...is that ok?

awkay:
Getting closer...just one code thing to test (no sente options) and a docstring fix.

src/main/fulcro/websockets.cljs
awkay: Also you said something about "auto", but I don't see it.

awkay: Nevermind!  I missed the call to merge and thought it was a separate option...

awkay:
Actually, I misread it. Only that docstring needs a fix, and I can do that.

PR #234, opened 2018-08-19
Needed to add this require in clojure side

currentoor:
Otherwise the consumer has to require this in clojure somewhere before they require `fulcro.client.primitives`.


awkay:
It needed a bit more than that.  Thanks for the attempt :)

PR #237, opened 2018-08-25
Remove basic-ui references

sagelliv:
The `basic-ui` does not reference anything previously in the book. I'm assuming the components are inlined in the `app.ui.root` namespace, but it should be made more explicit.


awkay:
Thanks

Issue #238, opened 2018-08-26
Transactions without quoting


thheller:
One thing that really bothers me about the transaction API is the quoting required to qualify the symbols.

I have an alternate proposal that would only require a tiny modification to the `defmutation` macro.

The use doesn't change
```
(defmutation the-mutation [...]
  (action [...]))
```

but when constructing a transaction the user could do

```
(fp/transact! this [(mut-ns/the-mutation {:params foo})])
;; instead of
(fp/transact! this `[(mut-ns/the-mutation {:params ~foo})])
```

`defmutation` could either just create `(defn the-mutation [params] (list 'some.mut-ns/the-mutation params))` or create a type/record instance which implements IFn.

```
(deftype TxDefinition [tx-id]
  IFn
  (-invoke [this]
    (list tx-id {}))
  (-invoke [this args]
    (list tx-id args)))

;; emitted by the macro (in addition to the multimethod)
(def the-mutation (TxDefinition. 'some.mut-ns/the-mutation))

(the-mutation {:params 1})
;; -> (some.mut-ns/the-mutation {:params 1})
```

The deftype/defrecord solution is slighty more extensible since it could also implement other protocols and take on more responsibility like spec-validating the passed params when called.

The quoted version would still work without modifications so it would be a non-breaking change.

This would also works nicely when the mutation impl is actually defined in other namespaces.

```
(ns my.app.api
  (:require [fulcro.api :refer (deftx)]))

(deftx the-mutation
  "docstring"
  param-spec)

(the-mutation {:params 1})
;; => (my.app.api/the-mutation {:params 1})

(ns my.app.client-impl
  (:require [my.app.api :as api] ...))

(defmutation api/the-mutation [...]
  ...)

(ns my.app.server-impl
  (require [my.app.api :as api] ...))

(defmutation api/the-mutation ...)
```

The main purpose is to get rid of the quoting for transactions. I can open a PR if interested.

awkay:
So we just started a `fulcro-incubator` project to play with ideas like these.  Wilker had some great ideas around additional "lifecycle" methods for transactions that he's already placed in there.   The current macro was created to solve really just one problem:  IDEs and tools don't deal well with multimethods.  With `defmutation` you get something that you can treat sorta like a function, and the tools can then navigate to (and show docstrings for).   It also helped reduce errors with returning a map (e.g. ppl would often write their actions in the body, which looks to work ok, but isn't right at all).

[...]

PR #239, opened 2018-08-29
add MutationHandle helper for callable mutations

awkay:


awkay:
As discussed in the issue #238 , this PR needs updated to check for a namespaced symbol, since this addition would otherwise cause a breaking change for some users.

awkay:
Added an implementatino to incubator

PR #243, opened 2018-09-21
Hide more macro-related code in :clj branches

thheller:
A lot of code is added to the CLJS build that can't be removed by Closure and is only relevant in CLJ parts of the code. Moved all the code into proper branches.

- ` cljs.analyzer` adds 85KB of unused code
- `fulcro.client.primitives` had another ~35KB of macro-only code

I only moved the worst offenders. There are a few more but most of them are spec-related.


awkay:
Very nice. Thank you!

PR #245, opened 2018-09-26
optimize component code generation with common prototype

thheller:
instead of emitting identical component lifecycle methods for
each component a common prototype is used instead

this saves about 1.5KB of :advanced optimized code per component


thheller:
I moved the macros to the bottom of the file because the common prototype is normal CLJS code and uses functions that must be declared before use.

The change is otherwise only to the `defui*` `set-react-proto!` fn and removing the `:defaults`. Otherwise no changes were done.

Issue #242, opened 2018-09-17
Activity Marker


awkay:
Currently only loads manipulate an activity marker.  Mutations are required to do their own manipulation.  This made sense initially because mutations are optimistic by default, and additional work must be done to deal with mutation return values, etc.  So, the intention was that you'd implement your own mutation activity process.

Wilker's proposal in incubator has some nice aspects in terms of giving an official lifecycle to a mutation, but it doesn't cover the "easy case" of just seeing that things are "in motion" on the network.

I was able to add an activity tracking into Fulcro reconciler's internal state with commit c291efebc2da12145f8d21b229f2ec4d23279338
which simply adds an internal map to the reconciler that can be accessed via `fulcro.client.impl.protocols/active-remotes`.

I'm considering what the reasonable approach is:

- Add a new key to top-level state that indicates "any network activity"
- Make the global loading activity marker really *mean* network activity, whcih is what many ppl expect.
- Add a `primitives/network-active?` function that returns true/false

The other motivating case at the moment is claudiu's attempt to do SSR with node.  He wants a way to get some kind of async notification that the network has become idle.  The fact that reconciler's state is an atom (not app state, reconciler state), means that a watch can be attached with the current scheme like so:

```
                :started-callback (fn [{:keys [reconciler] :as started-app}]
                                    (let [trigger (debounce (fn notify-idle-network [active-remotes]
                                                              (if (empty? active-remotes)
                                                                (js/console.log :NETWORK-IDLE))) 2)]
                                      (add-watch (:state reconciler) :active-remotes
                                        (fn [_ _ old-state new-state]
                                          (when (not= (:active-remotes old-state) (:active-remotes new-state))
                                            (trigger (:active-remotes new-state)))))))
```

The debounce is necessary because there will be rapid done/start edges if there are multiple requests queued. 2ms should be plenty, since really it's just giving up the CPU long enough for the next request to start.

This could be turned into an API call that would encompass this general facility, like this:

```
(defn notify-network-idle! [reconciler callback timeout] ...)
```

and the notification could even remove the watch for a single-shot notification.  The only use-case I see for this is node.js SSR, but it is a clearly useful thing for that.

I'd be interested in feedback.

awkay:
I've added a `prim/notify-network-idle!` in this commit:

ac1b3d66a2dc10cada995d3e404512786839d2f1

It codifies a single-shot way to ask for this.

claudiuapetrei:
[...]

PR #247, opened 2018-10-15
 #242 Add global network activity markers for remotes

claudiu-apetrei:


src/main/fulcro/client/primitives.cljc
```diff
@@ -2185,7 +2196,10 @@
       (p/queue! reconciler (into xs (remove symbol?) (keys v)))
       (when-not (empty? snds)
         (doseq [[remote _] snds]
+          (swap! (get-network-activity reconciler) assoc-in [remote :status] :active)
```
awkay: Were these really necessary, or would this be fixed by including the remote in the networking code's call of `queue!`?

awkay:
I would like to avoid modifying `transact*` if possible, but I realize that it may be necessary to prevent flicker. See my comment in code.

Also, please list out the scenarios you've tested.  Have you run the automated tests?

src/main/fulcro/client/primitives.cljc
claudiuapetrei: @awkay Not sure this is the best place. Added it there to make the load markers `sync`,  since it seems to be the common function between load & mutations, before they go on the async queue. The end result I was trying to get is:
```
(df/load reconciler ...)
(-> reconciler prim/app-state deref ::net/status :remote) ; = :active
```

claudiuapetrei:
I tested mostly in nodejs (headless app, load & renderString + the state), working now on setting up my app to test on the client also. 
Everything seems ok for mutations, mutations with load, load + parallel loads, multiple remotes. Still need to test ptransact scenarios.

Ran the test suite I have two specifications that are failing:
specification "Network payload processing (sequential networking)
specification "Raw API Request Handling (primary components for hand-building server)

If you think the approach is ok can look into fixing the tests (at a first glance looks like they failed because mocks need to be updated).

src/main/fulcro/client/primitives.cljc
awkay: OK, I see what you're going for.

Did it matter if the queue was with the remote or not?  Because you could just `conj` it onto the xs in the loop on the line before it.

awkay:
Yeah, I think the PR looks pretty good. I'm ok merging it, and I can add docs for it.  Take a look at the tests, and if you have trouble with them, let me know and I'll help.

claudiu-apetrei:
@awkay awesome. Tests are ok now, just needed to send a reconciler instance instead of `:reconciler` since now the networking uses it to get the network-activity atom.

PR #257, opened 2018-11-08
Use correct localhost server port 8080 in config/defaults.edn

oneness:
I really like what I have been seeing sofar. Thanks for the awesome
work. You might find me getting more involved with fulcro.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -238,7 +238,7 @@ and then add a `src/app/server.clj`:
 create a `resources/config/defaults.edn` that contains:
 
 ```
-{:port 3000}
+{:port 8080}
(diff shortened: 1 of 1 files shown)


awkay:
Thanks...why do you think this needs changed?

oneness:
Sorry to bother you again but did you mean people map localhost port 3000
to localhost port 8080?
Kasim Tuman

==========================================

Experience is more about knowing what should be avoided

than what should be done. -- Kasim
============================================

awkay:
Oh I see...sorry for the misunderstanding...yeah, this is fine.

PR #259, opened 2018-11-15
Add load-marker-default config option

claudiu-apetrei:


src/main/fulcro/client.cljc
```diff
@@ -164,6 +164,16 @@
                                              reconciler-options)
                        :networking         networking})))
 
+
+(defn new-fulcro-client
+  "Deprecated
+  "
+  [& {:as options}]                                         ; hack ?  [& options] returned IndexedSeq
+  (let [options-with-marker-default (update options :reconciler-options assoc :inplace-load-markers true)
```
awkay: perhaps should call it "load-marker-default", since we could let them set the default to a keyword or something as well.

src/main/fulcro/client/impl/protocols.cljc
```diff
@@ -13,6 +13,7 @@
   (get-id [this] "Get the unique ID of this reconciler. Used to resolve multiple running apps on a page. Defaults to the root class and a UUID.")
   (get-history [this] "Returns the current history atom")
   (get-network-activity [this] "Returns an atom containing network activity info.")
+  (get-config [this])
```
awkay: Lets not do that...it's just a keyword access

src/main/fulcro/client/data_fetch.cljc
```diff
@@ -184,12 +185,13 @@
             (prim/reconciler? app-or-comp-or-reconciler)
             #?(:cljs (implements? fc/FulcroApplication app-or-comp-or-reconciler)
                :clj  (satisfies? fc/FulcroApplication app-or-comp-or-reconciler)))]}
-   (let [config                  (merge {:marker true :parallel false :refresh [] :without #{}} config)
-         component-or-reconciler (if #?(:cljs (implements? fc/FulcroApplication app-or-comp-or-reconciler)
+   (let [component-or-reconciler (if #?(:cljs (implements? fc/FulcroApplication app-or-comp-or-reconciler)
```
awkay: need to update any docstrings affected

awkay:
so, you also need to update load-field, load-action, and load-field-action.

claudiu-apetrei:
@awkay Should have probably started with this, let me know if I got it right and how you want me to continue implementing this. Was thinking that:

The marker default should **not** be a global config. The default should be false, the only reason for adding it as a global config is to not have breaking changes for existing apps and provide a gradual migration option.

The global config should be a boolean. Allowing it as a keyword seem like shoot yourself in the foot material (parallel loads or having multiple remotes would compete for the keyword) 

I tried to introduce it into the codebase without affecting the public api to much. Thinking that in the future this global config should be removed so adding it to the docs and it having a nice name might encourage users to make use of it.

Let me know how you want me to approach these:
a) `Doc string should be updated to reflect new option.` Should I make it part of the public interface and document it ?
b) `perhaps should call it "load-marker-default", since we could let them set the default to a keyword or something as well.` Continue with :inplace-load-markers or rename it ?
c) `p/get-config` I need access to the config, p/get-config is a general option that might be useful in the future (personally not a big fan of hiding data from users) do you want me to create a specific method for getting the config value ?

awkay:
The protocol is not necessary in Fulcro (it was from Om Next) and will never be overridden.  The details are in the record and are already "public" in that respect.  

I'm trying to deprecate co-located load markers, period.  Being able to set the default to a keyword has collision possibilities, but it is "complete" in the sense that I won't then get further requests to add it when ppl want it.  Some ppl only do one load at a time, and won't want to be explicit everywhere, and I think co-located load markers were a bad idea to begin with.

The other option is just to force it to false, and make it so they always have to say where to put the load marker in load, and warn about the use of `true` in logs.

Clear goals:
- Make co-located markers a thing people don't use (get rid of true altogether, but with legacy support)
- Make it possible for them to say where their "default" load marker goes
- Make a new constructor that enforces the new rules, while keeping the old one to prevent breakage.

awkay:
I'm trying to remember why I recommend pulling in the whole marker table in the docs, too...seems like you could directly query for your load marker with the full ident: `[df/marker-table :marker-name]`...oh, I think it was because you cannot destructure that in props easily.  But you can do a get on props using that ident, so perhaps a helper function for that would be nice and clean.

```
(defn load-marker [props name] (get-in props [df/marker-table name]))
```

not sure exactly what to call it, or if that is necessary.

claudiu-apetrei:
Oky, got it now.  Thank you :) 
Will make the updates & implement it for load-field, load-action, and load-field-action also.

claudiu-apetrei:
Updated the commit, with some of the things we discussed (still have to apply other methods apart from df/load), a few questions came up:

- Added `:load-marker-default`, to the `make-fulcro-client` docs. Currently it's as a param for creating the client, but behind the scenes it goes to `reconciler-options`. Is it ok to leave it as an implementation detail or should the config be defined in the reconciler-options and documented there ?

- For the `get-config` did you mean to just add a method that exposes the keyword value ?

- For `true` to work from the old-client but not in the new. I added some metadata (:allow-load-marker-true) that is taken into consideration by the validate function. Added the check in `:pre`, the error is not the best, do you want me to try with spec (because of the meta, will probably have two specs and some conditional logic) or any other approach for the validation and error thrown ?

src/main/fulcro/client/data_fetch.cljc
```diff
@@ -296,13 +302,19 @@
...
         {:keys [without params remote post-mutation post-mutation-params fallback parallel refresh marker abort-id]
-         :or   {remote :remote refresh [] marker true}} params
-        env       (if (and (map? env-or-app-state) (contains? env-or-app-state :state))
-                    env-or-app-state
-                    {:state env-or-app-state})
-        state-map (some-> env :state deref)]
+         :or   {remote :remote refresh [] marker load-marker-default}} params
+        state-map           (some-> env :state deref)]
+    (when-not env?
+      (log/warn "Warn about the fact that state means no way to get the default ?")) ; TODO remember to update
```
awkay: This message should be 
```
(log/warn "load-field-action for field " field "was called with app state instead of env.  This is a deprecated usage and some features may not work when using it.")
```

src/main/fulcro/client/impl/protocols.cljc
```diff
@@ -13,6 +13,7 @@
   (get-id [this] "Get the unique ID of this reconciler. Used to resolve multiple running apps on a page. Defaults to the root class and a UUID.")
   (get-history [this] "Returns the current history atom")
   (get-network-activity [this] "Returns an atom containing network activity info.")
+  (get-load-marker-default [this])
```
awkay: Please do not try to make these protocols bigger.   I plan on removing them someday because they are absolutely senseless in the context of Fulcro and came from Om Next.  The reconciler is a defrecord, which means that any "field" you put in it is publicly accessible just like a map.  Please just use the keys. This is all internal implementation anyhow.

src/main/fulcro/client.cljc
```diff
@@ -73,7 +73,20 @@
 ; easier to access, so the user don't have to require a impl namespace to reference
 (def mutate app/write-entry-point)
 
-(defn new-fulcro-client
+
+(s/def ::load-marker-default (s/or :key keyword? :falsey (s/or :nil nil? :false false?)))
```
awkay: So, I'd drop the specs, and this validate function.  Let's let them set it to whatever the heck they want...it doesn't really matter (your spec isn't even technically right, since load marker names can be anything now). Drop the validate function...just make the "default if unspecified" be false.

src/main/fulcro/client.cljc
```diff
@@ -110,6 +123,8 @@
 
   `:shared` (optional). A map of arbitrary values to be shared across all components, accessible to them via (prim/shared this)
 
+  `:load-marker-default` (optional) Default is false, you can supply a keyword.
```
awkay: Load marker names can be anything. There are no restrictions anymore.

awkay:
> @awkay Added fulcro and developers guide documentation changes.
> 
> Don't know if there are any other places that have to be updated. Template + lein template I can submit a PR after the release to also update the fulcro version.

Yes, the templates should be updated, I don't care about other stuff...the old one still works.

> 
> All tests pass, but have a few doubts about:
> 
> * Don't know how to change the docs for http://book.fulcrologic.com/#_a_live_example_of_1_x_load_indicators

I'd still say they're supported, but drop the comment about them being the default. Probably should change the example code so that `:marker true` is passed.

> * Added no-params arity for make-fulcro-client
> * https://github.com/fulcrologic/fulcro/pull/259/files#diff-5698b9c89648f1e6d68fe01e3c279e60R97 not sure I corectly covered the usage of `~@args`

Right, just use the old function there...in fact, don't change internal uses to the new thing at all. The function is not going away, and can be used internally.  We're changing the docs, not removing the function.

> * There is a declare for `new-fulcro-client` it does not seem needed, should I add `make-fulcro-client` there also ?

Only want declare if it is actually needed...don't leave declares if they are not actually needed.

awkay:
Did you have more to do to this, or is it ready for a review/merge?

claudiuapetrei:
@awkay The PR is ready for review & merge.

awkay:
I'll review this soon.  I also want to make a few other defaults for this new constructor that modify some other default behaviors...this is a good chance to make all of them at once instead of having to keep renaming our client constructor function.

awkay:
I merged this in a combo with some changes of my own.

PR #261, opened 2018-11-16
Implement mutation/set-double!

bonega:
Actually uses parseFloat on js side, but is 64bit

### DIFF
--- a/src/main/fulcro/client/mutations.cljc
+++ b/src/main/fulcro/client/mutations.cljc
@@ -174,6 +174,23 @@
   (let [value (ensure-integer (if event (target-value event) value))]
     (set-value! component field value)))
 
+#?(:cljs
+   (defn- ensure-double [v]
+     (let [rv (js/parseFloat v)]
+       (if (js/isNaN rv) 0 rv)))
+   :clj
+   (defn- ensure-double [v] (Double/parseDouble v)))
+
+(defn set-double!
+  "Set the given double on the given `field` of a `component`. Allows same parameters as `set-string!`.
+
+   It is recommended you use this function only on UI-related data (e.g. data that is used for display purposes)
+   and write clear top-level transactions for anything else. Calls to this are compressed in history."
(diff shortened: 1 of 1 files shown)


src/main/fulcro/client/mutations.cljc
```diff
@@ -174,6 +174,26 @@
   (let [value (ensure-integer (if event (target-value event) value))]
     (set-value! component field value)))
 
+#?(:cljs
+   (defn- ensure-double
+     "Helper for set-integer!, use that instead. It is recommended you use this function only on UI-related
```
awkay: Might want to update (or remove) docstring

src/main/fulcro/client/mutations.cljc
```diff
@@ -174,6 +174,26 @@
...
+     data (e.g. data that is used for display purposes) and write clear top-level transactions for anything else."
+     [v]
+     (let [rv (js/parseFloat v)]
+       (if (js/isNaN rv) 0 rv)))
+   :clj
+   (defn- ensure-double [v] (Double/parseDouble v)))
```
awkay: This side will likely never get call, so this is nit-picky, but the JVM throws on bad number strings.

src/main/fulcro/client/mutations.cljc
bonega: I think we can do without the docstring

src/main/fulcro/client/mutations.cljc
bonega: Are you suggesting that we should catch and handle a bad value?
`set-integer!` does also throw on invalid strings.

awkay: nah, it's fine.  The clj side should never be running anyway.  Isomorphic is for first-frame rendering in a headless env...so no way for it to ever run.  Even writing a body is overkill other than making tests easier to write.

awkay:
Thanks

PR #266, opened 2018-12-15
Pre merge

### LINKED ISSUE #265: Pre merge hook proposal
In Fulcro we encourage and like to use all app state in a single atom. This adds some
excellent properties like making your app to reason about, but at the same time add some new
challenges to managing the data lifecycle.

There are a few key cases where new data enters a Fulcro local database:

1. At app initialization
2. Data loaded from a remote
3. User mutations adding new data

For cases 1 and 2 Fulcro provides some help with the `Initial State` protocol, for 2
It also provides the `post-mutation` in the load so you can alter things after they
get injected. For case 3 the Fulcro Incubator has the `create!` helper, which leverages
some of the `Initial State` to make easy the setup of a complex component hierarchy.

I like to demonstrate how those can be used today with a code example, and then
describe a new idea and how those could be rewritten once that is available.

== Case: Initializing to-many relations

Let's say we some kind of todo list where items are collapsible. After we load the data, we want the items to start expanded/collapsed depending on if the item is done. Done
items must start collapsed, pending must start open. Here is how we can write these
components:

```clojure
(fp/defsc Item
  [this {::keys [description]
         :ui/keys [expanded?]}]
  {:initial-state (fn [{::keys [done?]}] {:ui/expanded? (not done?)})
   :ident         [::item-id ::item-id]
   :query         [::item-id ::description ::done? :ui/expanded?]}
  (dom/div
    (if expanded?
      description)))

(def item (fp/factory Item {:keyfn :item-id}))

(fm/defmutation initialize-items [_]
  (action [{:keys [state ref]}]
    (swap! state
      (fn [state]
        (let [item-refs (get-in state (conj ref ::items))]
          (reduce
            (fn [state ref]
              (update-in state ref
                (fn [data]
                  (merge data (fp/get-initial-state Item data)))))
            state
            item-refs))))))

(fp/defsc ItemCollection
  [this {::keys [items]}]
  {:ident [::id ::id]
   :query [::id
           {::items (fp/get-query Item)}]}
  (dom/div
    (dom/button {:onClick #(df/load this (fp/get-ident this) ItemCollection
                             {:post-mutation `initialize-items})} "Load items")
    (mapv item items)))

(def item-collection (fp/factory ItemCollection))
```

This works but requires a significant amount of code to manually traverse the db and initialize
things, and this example is a single level thing, in real systems you may have deeper
dependencies, and as you go you add more and more code like that.

In the fulcro lifecycle for injecting new data, there is a part of it called `merge-component!`, which
does the component normalization and injection at DB. Merge component is nice because
it runs in all critical paths for data entering:

1. Its used to normalize the initial app state for the initial user data
2. Its used to normalize data from loads
3. Users can call it to add arbirtrary data

This seems a good strategic point to add a new hook we can call `:pre-merge`. This hook
will be called during the `merge-component!` process, since it already knows which component
it is merging it can call this with the `data tree` (not normalized) before it gets merged,
giving the component an entry point to modify it to add the UI needs.

This is what the code from the previous example would be using pre-merge:

```clojure
(fp/defsc Item
  [this {::keys   [description]
         :ui/keys [expanded?]}]
  {:pre-merge (fn [{::keys [done?] :as item}] 
                (assoc item :ui/expanded? (not done?)))
   :ident     [::item-id ::item-id]
   :query     [::item-id ::description ::done? :ui/expanded?]}
  (dom/div
    (if expanded?
      description)))

(def item (fp/factory Item {:keyfn :item-id}))

(fp/defsc ItemCollection
  [this {::keys [items]}]
  {:ident [::id ::id]
   :query [::id
           {::items (fp/get-query Item)}]}
  (dom/div
    (dom/button {:onClick #(df/load this (fp/get-ident this) ItemCollection)} "Load items")
    (mapv item items)))

(def item-collection (fp/factory ItemCollection))
```

In the `:pre-merge` hook you receive the data tree, modify as you see fit just before it 
gets normalized and injected in the local DB.

This removes the need for the `post-mutation` entirely in this case. During `post-mutation`
you have to deal with the normalized DB, in `pre-merge` you get the tree, making it easier
to do even complex modifications with simpler code. 

I like to give another example, this time imagine you are rendering another list with
collapsible items. But this time we want the item with the higher score to start expanded,
in case of ties all the ties should start expanded, lets see how we could write this
using `:pre-merge`:

```clojure
(fp/defsc Score
  [this {::keys []}]
  {:pre-merge (fn [score] (merge {:ui/expanded? false} score))
   :ident     [::score-id ::score-id]
   :query     [::score-id ::points :ui/expanded?]})

(def score (fp/factory Score {:keyfn ::score-id}))

(fp/defsc Scoreboard
  [this {::keys []}]
  {:pre-merge (fn [{::keys [scores] :as scoreboard}]
                (let [high-score (apply max (map ::points score))
                      scores     (mapv
                                   (fn [{::keys [points] :as score}]
                                     (assoc score :ui/expanded? (= points high-score)))
                                   scores)]
                  (assoc scoreboard ::scores scores)))
   :ident     [::scoreboard-id ::scoreboard-id]
   :query     [::scoreboard-id
               {::scores (fp/get-query Score)}]})

(def scoreboard (fp/factory Scoreboard))
```

Merge component will run from the top to the bottom, so the parent gets called first
and it can change it's children without having to deal with normalized data. Also
note on the `Score` can ensure `:ui/expanded?` is initialized even if the parent hadn't
done it.

This makes for a simpler code to initialized merged data, and this also composes better
than `post-mutations` since its defined at the component level and adds up as components
are hooked on each other.

This is not a replacement for `post-mutations`, the `post-mutations` is a powerful tool
were the entire DB is available to you and enables updates of any complexity; the
`pre-merge` is more limited but can cover a lot of cases, reducing the need for `post-mutations`
and enabling better composition of component data merge.

I like to hear what others think about it, if you have any suggestions or critics please
let's discuss on this issue.

wilkerlucio:
Initial implementation proposal.

- [x] Add IPreMerge protocol
- [x] Add `:pre-merge` syntax to `defsc`
- [x] Apply pre-merge during normalization when provided
- [x] Make sure compiler errors from defsc are good when you make a mistake.
- [x] Write documentation on book
- [x] Cover previous cases of tree->db with tests (were missing before)
- [ ] Experiment on real apps

Closes #265

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ small and tractable.
 
 2.8.0
 ------
-- Added pre-merge support
+- Add pre-merge support
 
 2.7.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ small and tractable.
 - (inc y) = Intrusive change that should not break anything, but should be heavily tested.
 - (inc z) = additions, bug fixes, etc.
 
-2.7.0
+2.6.19
 ------
 - Added pre-merge support
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+2.8.0
+------
+- Add pre-merge support
+
 2.7.2
 -----
 - Fixed deprecated websockets to work with new sente.
@@ -5,10 +9,6 @@
 2.7.1
 -----
 - Minor fix to i18n
-
-2.8.0
-------
-- Add pre-merge support
 
 2.7.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ small and tractable.
 - (inc y) = Intrusive change that should not break anything, but should be heavily tested.
 - (inc z) = additions, bug fixes, etc.
 
+2.7.0
+------
+- Added pre-merge support
+
 2.6.18
 ------
 - Added a bunch of missing SVG tags to DOM.
--- a/DevelopersGuide.adoc
+++ b/DevelopersGuide.adoc
@@ -2642,6 +2642,7 @@ With all hook up here is the demo:
 ++++
 <div class="example" id="pre-merge-postmutations"> </div>
 <a class="source-toggle" href="javascript:void(0)" onClick="$('.pre-merge-postmutations').toggle()">Show/Hide Source</a>
+<br><br>
 ++++
 
 [source,clojure,role="pre-merge-postmutations source"]
@@ -2667,6 +2668,7 @@ Now let's see in case of a to-many, this is what our new mutation looks like:
 ++++
 <div class="example" id="pre-merge-postmutations-many"> </div>
 <a class="source-toggle" href="javascript:void(0)" onClick="$('.pre-merge-postmutations-many').toggle()">Show/Hide Source</a>
+<br><br>
 ++++
 
 [source,clojure,role="pre-merge-postmutations-many source"]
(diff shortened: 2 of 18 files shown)


awkay:
How's this going?

wilkerlucio:
Hello, the code is done I think, I have a snapshot version that's running in production with the changes since last week, appears to have no regressions so far, I'm using the pre-merge in some parts of the code but users are not using that yet. 

On the tech side, it might worth the changes that happened since the initial proposal. After some experimentation I realized that the new tree is not enough, we also need the current data, otherwise, the pre-merge would override the data every time. This imposes another decision there is if the current data should be provided normalized or denormalized, after some discussion with @awkay we agreed the normalize is better, and for completeness you also receive the full app state map (not the atom, read-only), considering the many things now the input for the protocol is a map, so the users can pull only the pieces they want, and keeps the API future-proof (we can always add more things to the map).

This is what the code looks like using the new interface, updating the code from the proposal example:

```clojure
(fp/defsc Item
  [this {::keys   [description]
         :ui/keys [expanded?]}]
  {:pre-merge (fn [{:keys [data-tree current-normalized]}]
                (let [{::keys [done?] :as item} data-tree]
                  (merge
                    {:ui/expanded? (not done?)}
                    current-normalized
                    data-tree)))
   :ident     [::item-id ::item-id]
   :query     [::item-id ::description ::done? :ui/expanded?]}
  (dom/div
    (if expanded?
      description)))
```

Using this merge with 3 elements (defaults, current data, new data) makes a good default for most cases while providing very fine control to the user.

I plan to finish the tests and docs by the jan first week when I get back home.

Please let me know if you still wanna know any more details. Thanks.

PR #270, opened 2019-01-09
Filter deprecated translations

tcoenraad:
It appeared to us, that deprecating translations would cause warnings after i18n extraction and merger. On page load multiple lines of `Unexpected input --> <--` were dropped, originating here https://github.com/fulcrologic/fulcro/blob/33bae48ef2830137059d74066ef61396583821f9/src/main/fulcro/gettext.clj#L45.

This PR fixes these warnings by removing all superfluous lines that would be blank after comment stripping, so they are not considered at all when creating the translation map.

### DIFF
--- a/src/main/fulcro/gettext.clj
+++ b/src/main/fulcro/gettext.clj
@@ -24,7 +24,7 @@
 
 (defn get-blocks
   [resource]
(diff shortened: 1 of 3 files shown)


awkay:
Sounds good to me.  Glad to hear people are using it.

Issue #271, opened 2019-01-11
dom/wrap-form-element misbehaves when form tag is nested in other tags


j1mr10rd4n:
React emits the following warning
```
Warning: A component is changing a controlled input of type text to be uncontrolled
```

when a component that is wrapped by [dom/wrap-form-element](https://github.com/fulcrologic/fulcro/blob/ed9cc740ce5a11f9c4eeb2fa381e82599e83f5a0/src/main/fulcro/client/dom.cljs#L112) and has its input tag nested inside other tags is first changed, for example:
```
<div class="fancy-labels-etc">
  <input type="text" value="Hello" onChange=(function) >
</div>
```
(the semantic-ui-react [Input](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Input/Input.js) element has this structure).

This is a side effect of the assumption in [line 135](https://github.com/fulcrologic/fulcro/blob/ed9cc740ce5a11f9c4eeb2fa381e82599e83f5a0/src/main/fulcro/client/dom.cljs#L135) that the first node returned by `js/ReactDOM.findDOMNode` is the node containing the value.

A minimal project to reproduce the bug can be found at [j1mr10rd4n/fulcro-wrap-form-element-bug](https://github.com/j1mr10rd4n/fulcro-wrap-form-element-bug).

Screen capture demonstrating the issue:
![wrap-form-input-bug-demo](https://user-images.githubusercontent.com/946427/51024324-13737480-153e-11e9-96e6-a47cb323a879.gif)

awkay:
Thanks!

PR #272, opened 2019-01-11
Fix dom/wrap-form-element bug #271

j1mr10rd4n:
This PR addresses #271 by drilling down through any nested tags that may be present around the wrapped form element to establish its value. Tests are included.

Introduces a test dependency on [enzyme](https://airbnb.io/enzyme/).

Some hanging :focused markers are also removed.


awkay:
Hi Jim,

Thanks.  I take it you've manually tested this a reasonable amount both with controlled components that use both Fulcro state and component-local state?  I don't see any reason why it would break things, but this would be quite annoying to users if it did.

I guess they can always make a bug report and downgrade if there is a problem.

Thanks!

PR #274, opened 2019-01-11
Transform and pass response headers through client middleware

j1mr10rd4n:
This PR enhances the client middleware slightly so that it includes the xhrio response headers as a clojure map.

One of my projects needed this functionality for reading custom authorization headers sent from a server, for example.


awkay:
Sure. Seems harmless.

Issue #295, opened 2019-02-14
Form state gives incorrect form-delta for unrequired subforms


TimovanderKamp:
Lets say this is your form setup:

```
(defsc Author [_ props]
  {:query [:id
           :name
           fs/form-config-join]
   :form-fields #{:id :name}
   :ident [:author/by-id :id]})

(defsc TodoList [_ props]
  {:query [:id
           :title
           {:author (prim/query Author)}
           fs/form-config-join]
   :form-fields #{:id :title :author}
   :ident [:todo/by-id :id]}
  (dom/div (str (fs/dirty-fields props true))))
```

If you add form config to this initial form state: `{:id 1 :title "My todolist"}`. You will get the following dirty-fields without changing anything: `{:author {:before nil :after []}}`

awkay:
Sorry, I'm not ignoring you...just busy.

TimovanderKamp:
No worries

awkay:
So, this is a little sensitive, since I don't actually know the arity of the relation.  Technically I think the correct fix is to just omit the field altogether.

awkay:
Released in 2.8.2-SNAPSHOT.

Thanks

TimovanderKamp:
Thanks for merging!

PR #296, opened 2019-02-14
Do not mark a subform as changed if it was initialized with a nil value

TimovanderKamp:
Resolves the issue addressed in #295


src/main/fulcro/ui/form_state.cljc
```diff
@@ -479,7 +479,7 @@
                                        current-value (cond
                                                        (map? items) (subform-ident k items)
                                                        (vector? items) (mapv #(subform-ident k %) items)
-                                                       :else [])
+                                                       :else items)
```
awkay: This is a keep, I think the right thing to do is return nil, so it is left out.

awkay: Ah, I see, your change is correct I believe.

PR #298, opened 2019-03-02
Default url value in fulcro-http-remote

souenzzo:


awkay:
Thanks...I keep meaning to fix that :)

PR #314, opened 2019-04-23
declaring the tempid default tag

souenzzo:


src/main/fulcro/tempid.cljc
```diff
@@ -1,6 +1,8 @@
 (ns fulcro.tempid
   #?(:clj (:import [java.io Writer])))
 
+(def tag "fulcro/tempid")
```
awkay: If you want to add this, I'd prefer it be authoritative...that is, used within the source instead of just declared.

souenzzo: done! :)

souenzzo:
Now I'm using the tag from tempid in transit ns's too.

awkay:
Cool, thanks

PR #327, opened 2019-09-06
Fixed incorrect docs

currentoor:


src/main/com/fulcrologic/fulcro/ui_state_machines.cljc
```diff
@@ -1059,10 +1059,10 @@
 
   `::comp/component-class` - The defsc name of the component to use for normalization and query. Only needed if the
     actor was not declared using a Fulcro component or component class.
-  `::uism/post-event`:: An event to send when the load is done (instead of calling a mutation)
-  `::uism/post-event-params`:: Extra parameters to send as event-data on the post-event.
-  `::uism/fallback-event`:: The event to send if the load triggers a fallback.
-  `::uism/fallback-event-params`:: Extra parameters to send as event-data on a fallback.
+  `::uism/ok-event`:: An event to send when the load is done (instead of calling a mutation)
```
awkay: you sure??? I thought I kept the old events on load...

currentoor: 100% this docstring is the only place that mentions `::uism/post-event` etc. Also both loads call this function
https://github.com/fulcrologic/fulcro/blob/develop/src/main/com/fulcrologic/fulcro/ui_state_machines.cljc#L1000

If you recall I suggested loads and remote mutation mutations have the same post event names to unify the API. I'm glad you did this :smile:

awkay: Could you also check the PORTING guide file and Dev Guide???

currentoor: Oh right I forgot about those, I'll update.

currentoor: Both of those places were already uptodate.

PR #331, opened 2019-09-07
Small improvements

green-coder:
Replaced some usage of the reduce function with transducers. The goal is to reduce garbage collection and improve performances.

Notes:
- I did not test if the code is correct or if it does improve performance.
- Feel free to edit the commit or take smaller parts if you see fit.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
+++ b/src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
@@ -17,13 +17,11 @@
   "Checks the given `idents` and returns a subset of them where the data they refer to has changed
    between `old-state` and `new-state`."
   [old-state new-state idents]
-  (reduce
-    (fn [result ident]
-      (if (identical? (get-in old-state ident) (get-in new-state ident))
-        result
-        (cons ident result)))
-    (list)
-    idents))
+  (into []
+        (filter (fn [ident]
+                  (not (identical? (get-in old-state ident)
+                                   (get-in new-state ident)))))
+        idents))
 
 (defn render-component!
   "Uses the component's query and the current application state to query for the current value of that component's
@@ -79,13 +77,10 @@
     (let [{:com.fulcrologic.fulcro.application/keys [runtime-atom]} app
           {:com.fulcrologic.fulcro.application/keys [indexes]} @runtime-atom
           {:keys [prop->classes class->components]} indexes]
-      (reduce
-        (fn [result prop]
-          (let [classes    (prop->classes prop)
-                components (reduce #(into %1 (class->components %2)) #{} classes)]
-            (into result components)))
-        #{}
-        property-set))))
+      (into #{}
+            (comp (mapcat prop->classes)
+                  (mapcat class->components))
+            property-set))))
 
 (defn render-stale-components!
   "This function tracks the state of the app at the time of prior render in the app's runtime-atom. It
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
```diff
@@ -17,13 +17,11 @@
   "Checks the given `idents` and returns a subset of them where the data they refer to has changed
    between `old-state` and `new-state`."
   [old-state new-state idents]
-  (reduce
-    (fn [result ident]
-      (if (identical? (get-in old-state ident) (get-in new-state ident))
-        result
-        (cons ident result)))
-    (list)
-    idents))
```
awkay: This was written with lists and cons *for* speed. Cons is making a linked list, and adding to the head is very very fast. There is no intermediate data here, and your impl is probably slower.

src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
```diff
@@ -79,13 +77,10 @@
...
-        (fn [result prop]
-          (let [classes    (prop->classes prop)
-                components (reduce #(into %1 (class->components %2)) #{} classes)]
-            (into result components)))
-        #{}
-        property-set))))
+      (into #{}
+            (comp (mapcat prop->classes)
+                  (mapcat class->components))
+            property-set))))
```
awkay: This seems an improvement. Realize that property-set is usually one thing...maybe a few. Rarely more than 3.

src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
green-coder: You might be right. I was tempted to replace it `(list)` with `#{}` but that would mean changing the API.

src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
```diff
@@ -108,15 +103,12 @@
...
-                                     (conj acc p)
-                                     acc))
-                                 #{}
-                                 linked-props)
+            dirty-linked-props (into #{}
+                                     (filter (fn [p]
+                                               (not (identical?
+                                                      (get state-map p)
+                                                      (get last-rendered-state p)))))
+                                     linked-props)
```
awkay: This is technically an improvement, but again realize that the number of linked props should be relatively small for any given render (users don't change much per transaction usually).  To be honest, I'm not even sure the ident-optimized render ends up "winning" over plain keyframe in a lot of cases...esp if there *were* a lot of dirty props.  It is entirely possible that the *real* optimization in this system is a heuristic that can guess when a keyframe render is going to win before doing all of this other work.

green-coder:
When I saw the implementation of `render-stale-components!`, it feels like a lot of the code in the `let` expression could be expressed as a composed transducer, removing some intermediate data starage and dedup. If you want, I can give it a try.

src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
green-coder: I am still learning Fulcro at the moment. Once I am done reading the full documentation and actually use it, I would be happy to contribute to that type of profiling.

src/main/com/fulcrologic/fulcro/rendering/ident_optimized_render.cljc
awkay: I just did a manual test:

```
(time (do 
        (reduce (fn [r v]
                 (cons v r)) (list) (range 1 100000))
        :ok))
"Elapsed time: 1.98474 msecs"
```

vs yours:

```
(time (do 
        (into [] (range 1 100000))
        :ok))
"Elapsed time: 2.073323 msecs"
```

Notice that the per-item insertion time for both is around 20ns each...and I'd say the usualy number of idents on-screen in an app might average about 100-1000...so 20 microseconds.

Premature optimization for sure...and it is not really any different.

awkay:
I'm closing this for now.  If you get around to testing it then you can re-open a PR.

PR #333, opened 2019-09-25
Fix dynamic routing deferred handling

mitchelkuijpers:
If you use a dr/deferred there is a timing issue which can get you in a
forever loading state and then the error state. It happens as follows:

You load a route with deferred-route, which does a network call which
takes around 60/70 ms. If the call succeeds you call dr/target-ready.
It triggers the routed state and starts canceling all timers, but the
timeout timer still manages to trigger the waiting! event. It goes to
:routed and then it actually triggers the waiting! event which has a
single handler which tries to route to a empty route, it then goes to
timeout and then the failed state.

For further clarification this is what we see happening in the state
machine.
```
 --------------------------------------------------------------------
|      State      |    Event     |           Remarks                 |
|  :routed        |    :route!   | It sees it is a deferred route    |
|                 |              | and starts delay and error timers |
|  :deferred      |              |                                   |
|  :deferred      |    :ready!   | when this is happening the        |
|                 |              | the timeout timer triggers it gets|
|                 |              | canceled but the callback still   |
|                 |              | triggers                          |
|  :routed        |              | this is where we want it to stop  |
|  :routed        |  :waiting!   | this caused by the timeout!       |
|                 |              | which got triggered even though it|
|                 |              | got canceled. it triggers the     |
|                 |              | default handler it tries to route |
|                 |              | to a deferred route without a     |
|                 |              | target                            |
|  :deferred      |              |                                   |
|  :deferred      |  :waiting!   |                                   |
|  :pending       |              |                                   |
|  :pending       |  :timeout!   |                                   |
|  :failed        |              |                                   |
```
We fixed this by making the routes state listen explicitly for a route!
event so it does not go haywire when a timer still triggers even though
it get's canceled.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -292,7 +292,7 @@
                 :ready! {::uism/target-state :routed
                          ::uism/handler      ready-handler}}}
 
-    :routed   {::uism/handler route-handler}}})
+    :routed   {::uism/events {:route! {::uism/handler route-handler}}}}})
 
 ;; TODO: This algorithm is repeated in more than one place in slightly different forms...refactor it.
 (defn proposed-new-path [this-or-app relative-class-or-instance new-route]


awkay:
Cool, thanks for looking into this...so this is an issue with js not honoring a timer cancel? Rather, I bet the timer fired before being cancelled and scheduled the event, so then our call to cancel is a no-op because the js execution of the timer is already in the (js) UI thread's queue?  Either way, I agree that your change will fix it.

mitchelkuijpers:
Yes i think the call to cancel is no-op and the trigger is already scheduled :+1:

PR #337, opened 2019-10-05
[feat] introduces a new namespace for db-helpers with basic utils

abhi18av:
This PR adds a new namespace to the `algorithms`, namely `com.fulcrologic.fulcro.algorithms.normalized-state-helpers` as well as the corresponding test suite `com.fulcrologic.fulcro.algorithms.normalized-state-helpers-spec`.

A brief index of the utility functions is as follows

- `integrate-ident*`
- `remove-ident*`
- `tree-path->db-path`
- `get-in`
- `ui->props`
- `remove-entity*`
- `remove-edge*`
- `sort-idents-by`
- `update-caller!`
- `update-caller-in!`
- `swap!->`

For reference, this PR takes some code from the [WIP] https://github.com/abhi18av/fulcro/pull/2  and https://github.com/abhi18av/fulcro/pull/4 in order to get the ball rolling.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
@@ -1,0 +1,309 @@
+(ns com.fulcrologic.fulcro.algorithms.normalized-state-helpers
+  "Functions that can be used against a normalized Fulcro state database."
+  #?(:cljs (:require-macros com.fulcrologic.fulcro.algorithms.normalized-state-helpers))
+  (:refer-clojure :exclude [get-in])
+  (:require
+    [clojure.set :as set]
+    [com.fulcrologic.fulcro.components :as comp :refer [defsc]]
+    #?(:clj  [com.fulcrologic.fulcro.dom-server :as dom]
+       :cljs [com.fulcrologic.fulcro.dom :as dom])
+    [edn-query-language.core :as eql]
+    [clojure.spec.alpha :as s]
+    [ghostwheel.core :refer [>defn =>]]
+    [com.fulcrologic.fulcro.algorithms.data-targeting :as targeting]
+    [com.fulcrologic.fulcro.algorithms.denormalize :as fdn]
+    [com.fulcrologic.fulcro.algorithms.merge :as merge]
+    [com.fulcrologic.fulcro.components :as comp]))
+
+(def integrate-ident*
+  "[state ident & named-parameters]
+
+  Integrate an ident into any number of places in the app state. This function is safe to use within mutation
+  implementations as a general helper function.
+
+  The named parameters can be specified any number of times. They are:
+
+  - append:  A vector (path) to a list in your app state where this new object's ident should be appended. Will not append
+  the ident if that ident is already in the list.
+  - prepend: A vector (path) to a list in your app state where this new object's ident should be prepended. Will not place
+  the ident if that ident is already in the list.
+  - replace: A vector (path) to a specific location in app-state where this object's ident should be placed. Can target a to-one or to-many.
+   If the target is a vector element then that element must already exist in the vector.
+
+  NOTE: `ident` does not have to be an ident if you want to place denormalized data.  It can really be anything.
+
+  Returns the updated state map."
+  targeting/integrate-ident*)
+
+(def remove-ident*
+  " [state-map ident path-to-idents]
+
+  Removes an ident, if it exists, from a list of idents in app state. This
+  function is safe to use within mutations."
+  merge/remove-ident*)
+
+
+(>defn tree-path->db-path
+  "Convert a 'denormalized' path into a normalized one by walking the path in state and honoring ident-based edges.
+
+  For example, one might find this to be true for a normalized db:
+
+  ```
+  state => {:person/id {1 {:person/id 1 :person/spouse [:person/id 3]}
+                        3 {:person/id 3 :person/first-name ...}}}
+
+  (tree-path->db-path state [:person/id 1 :person/spouse :person/first-name])
+  => [:person/id 3 :person/first-name]
+  ```
+  "
+  ([state path]
+   [map? vector? => vector?]
+   (loop [[h & t] path
+          new-path []]
+     (if h
+       (let [np (conj new-path h)
+             c (clojure.core/get-in state np)]
+         (if (eql/ident? c)
+           (recur t c)
+           (recur t (conj new-path h))))
+       (if (not= path new-path)
+         new-path
+         path)))))
+
+
+(>defn get-in
+  "Just like clojure.core/get-in, but if an element of the path is an ident it will follow the ident instead."
+  ([state-map path]
+   [map? vector? => any?]
+   (get-in state-map path nil))
+
+  ([state-map path not-found]
+   [map? vector? any? => any?]
+   (clojure.core/get-in state-map (tree-path->db-path state-map path) not-found)))
+
+
+(defn ui->props
+  "Obtain a tree of props for a UI instance from the current application state. Useful in mutations where you want
+  to denormalize an entity from the state database. `this` can often be obtained from the mutation `env` at the
+  `:component` key."
+  ([this]
+   (ui->props (comp/component->state-map this) (comp/react-type this) (comp/get-ident this)))
+  ([state-map component-class ident]
+   (fdn/db->tree (comp/get-query component-class state-map) (get-in state-map ident) state-map)))
+
+
+(defn- dissoc-in
+  "Dissociates an entry from a nested associative structure returning a new
+   nested structure. keys is a sequence of keys. Any empty maps that result
+   will not be present in the new structure."
+  [m [k & ks :as keys]]
+  (if ks
+    (if-let [nextmap (get m k)]
+      (let [newmap (dissoc-in nextmap ks)]
+        (if (seq newmap)
+          (assoc m k newmap)
+          (dissoc m k)))
+      m)
+    (dissoc m k)))
+
+
+(>defn remove-entity*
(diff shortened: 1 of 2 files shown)


abhi18av:
Added the `remove-entity*`, needed to change the implementation a bit along with the tests.

abhi18av:
As of now, the initial batch of util functions has been added along with their tests.

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,303 @@
...
+                           (filter (fn [a-path]
+                                     (let [vl (clojure.core/get-in state a-path)]
+                                       (if (coll? vl)
+                                         (or
+                                           (some #{ident} vl)
+                                           (= ident vl))
+                                         (= ident (take 2 a-path)))))
+                                   (normalized-paths state)))
+
+         remove-ident-from-table (fn [state1 table]
```
awkay: This can't be right, not even using `table`

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,303 @@
...
+   (remove-entity* state-map ident #{}))
+
+  ([state1 ident cascade]
+   [map? eql/ident? (s/coll-of keyword? :kind set?) => map?]
+   (let [tables (keep (fn [k]
+                        (let [candidate (get state1 k)]
+                          (when (and (map? candidate) (every? map? (vals candidate)))
+                            k))) (keys state1))
+
+         non-tables (keep (fn [k]
```
awkay: This should end up as a sequence of keys. Wrapping each in a vector here seems wrong.

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,303 @@
...
+                                         (or
+                                           (some #{ident} vl)
+                                           (= ident vl))
+                                         (= ident (take 2 a-path)))))
+                                   (normalized-paths state)))
+
+         remove-ident-from-table (fn [state1 table]
+                                   (reduce
+                                     remove-idents-at-path
+                                     state1
```
awkay: Probably a good idea to use an alt name for `state1` here just so it isn't confused by reader with f arg

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,303 @@
...
+         non-tables (keep (fn [k]
+                            (let [candidate (get state1 k)]
+                              (when (vector? candidate)
+                                [k])))
+                          (keys state1))
+
+         remove-idents-at-path (fn [state1 path]
+                                 (let [v (clojure.core/get-in state1 path)]
+                                   (cond
+                                     (int? v) state1
```
awkay: Why isn't `int?` just the `else` case?

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,303 @@
...
+                                        tables)
+                                ;; remove the top-level entity
+                                (dissoc-in ident))
+
+         target-entity (get-in state1 ident)
+
+         cascaded-idents (fn [original-state target-entity cascade]
+                           (map
+                             (fn [x] (clojure.core/get-in original-state
+                                                          (conj [(first ident) ((first ident) target-entity)] x)))
```
awkay: A `let` around this would make it a lot clearer. `(let [table-key (first ident)] ...)`

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
abhi18av: Actually, this is a reducing function and the `table` parameter is passed while computing `state-without-entity` on Line 179

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
abhi18av: In order to use the `concat` on Line 174, I thought that maybe it's better to put this piece of logic here itself, shall I move it to L-174 itself ( which might end up doing too many things in a single form ). What do you suggest as an alternative?

abhi18av: Hmm, I found a better way perhaps via set theoretic difference. This way, I can build upon the logic in `table`

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
abhi18av: Yeah, this does seem like a patchwork solution - working on it.

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
abhi18av: Oh, now I got what you meant 😅 

I'm on it.

abhi18av: I've done some refactoring and it turns out that `tables/non-tables` were redundant since we already had `normalized-paths`. 

Have pushed the updated version.

abhi18av:
I have definitely become curious in generative testing :)

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
```diff
@@ -0,0 +1,309 @@
...
+                           (reduce (fn [new-state ident] (remove-entity* new-state ident cascade)) state edge)
+                           (remove-entity* state edge cascade)))
+                       state-without-entity
+                       cascaded-idents)]
+
+     final-state)))
+
+
+(>defn remove-edge*
+  "Remove the given edge at the given path. Also scans all tables and removes any to-one or to-many idents that are
```
awkay: Now that I'm looking at this one I'm not loving it.

I think this needs additional option(s).  The use-case in applications is that you're writing a mutation where
you know the edge name, and you'd like to unlink it from the graph; however, it is often the case that you do *not* want to actually cause the removal of the entity itself (that's remove-entity*).  For convenience, this function *becomes* a convenience interface for `remove-entity*`.  Technically what is at the edge is one of 3 things:

1. A map or scalar: ignore the request
2. A single ident: Two choices. 1. Remove the edge (simple dissoc). 2. Call `remove-entity*` with that ident.
3. Many idents: Reduction using rule in (2).

Thus, being able to tell it which behavior (dissoc vs. remove-entity), and *then if* it is remove entity, where to cascade. 

In real mutations you're often going to want to use `ref` (remove edge from "this"):

```
(-> s
  (remove-edge* (conj ref :person/children)))
```

or perhaps some top-level edge `(remove-edge* state-map [:root-key])`.

Cascading/recursion is rarer in these cases.

Perhaps this?

`(remove-edge* s path)` - pretty much dissoc-in. No entities are removed, just the edge.
`(remove-edge* s path #{})` - Removes the entity (and all refs to it), but since cascade set is empty goes no further.
`(remove-edge* s path #{:edge})` - Follows the given edges as well as removing the entity.

Thoughts?

wilkerlucio: one question, since those are public functions, why don't we remove the `*` from the end of the name? some old had because were internal, but looks to me for a public one, what you think?

awkay:
Our naming convention is a `*` suffix if it is meant to operate on the
state map (not atom)...has nothing to do with internal vs external.

src/main/com/fulcrologic/fulcro/algorithms/normalized_state_helpers.cljc
awkay: Our naming convention is a `*` suffix if it is meant to operate on the
state map (not atom)...has nothing to do with internal vs external. If we wanted to abandon that naming convention it leave us in a state where a lot of the older ones had the suffix and the newer ones do not, not to mention the stuff I've written about this in the book.  

That said, I realize that it is a non-standard naming convention. I'd be willing to consider a proposal to drop the `*`, and add aliases for all of the existing ones and dropping the `*`.

wilkerlucio: I'm down for that (dropping `*`), better to have the ones to atom to have `!`, so we follow regular conventions for Clojure

abhi18av: My experience is, of course, quite limited in contrast to both of you. The only comment I'd add is that I agree sticking to the standard clojure naming convention however, the `aliases` part gives me a hiccup for sure. It might make the API surface seem unnecessarily large which might end up confounding people for no reason.

abhi18av: > `(remove-edge* s path)` - pretty much dissoc-in. No entities are removed, just the edge.

Mm, if it is a simple `dissoc-in` ( which is an internal function util as of now), why not expose it and let the users use it explicitely

wilkerlucio: 👍  about leaving `dissoc-in` public

awkay:
I'm going to merge this, and then touch it up a bit. I think some of the names/docstrings should probably be tweaked a bit before release.

abhi18av:
Thanks 🌟

PR #339, opened 2019-10-11
Fix update query to compose with query-transform-default

wilkerlucio:


src/main/com/fulcrologic/fulcro/data_fetch.cljc
```diff
@@ -269,6 +269,9 @@
                                          (cond-> {:marker load-marker-default :parallel false :refresh [] :without #{}}
                                            query-transform-default (assoc :update-query query-transform-default))
                                          config)
+         config        (cond-> config
+                         (and (:update-query config) query-transform-default)
+                         (assoc :update-query (comp query-transform-default (:update-query config))))
```
awkay: I'm not sure this is what I want here.  For loads, the query transform default is just that: what to do when you don't supply one.  I think the global-eql-transform is the one that can manipulate it in a general way.  Also, won't this possibly mix it in twice since it is added as the :update-query on line 270?

awkay:
I don't think this is a wanted PR...closing.  We can reconsider how the transforms work, but mixing it here isn't right I don't think

PR #341, opened 2019-10-16
children function always returns a collection as it did in fulcro 2

timeyyy:


awkay:
Thanks!

awkay:
In 3.0.5 (SNAPSHOT for now)

PR #342, opened 2019-10-21
[docs] improve docstring coverage for the public API 

abhi18av:
I noticed some of the functions from the public API don't have the associated `docstrings` - this PR is an attempt to improve the docstring coverage.

I've made a draft attempt here, please let me know 

- if the added docstrings are accurate description of the intent for that API
- if the API is intended to be used internally only, which ought to reflect in the docstring

### DIFF
--- a/src/main/com/fulcrologic/fulcro/ui_state_machines.cljc
+++ b/src/main/com/fulcrologic/fulcro/ui_state_machines.cljc
@@ -20,7 +20,10 @@
 (declare asm-value trigger-state-machine-event! apply-action)
 (def mutation-delegate (m/->Mutation `mutation-delegate))
 
-(defn set-js-timeout! [f tm] (sched/defer f tm))
+(defn set-js-timeout!
+  "Schedules a `js/setTimeout` after the given time period `tm` (in milli-seconds)."
+  [f tm]
+  (sched/defer f tm))
 
 (defn clear-js-timeout! [timer]
   #?(:clj  (when timer (timer))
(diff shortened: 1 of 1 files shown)


awkay:
Hey there.  So, I appreciate the effort to improve docstrings.  My guidelines for these are to try to be a bit more helpful.  The dostring should really do more than the function name itself already does (e.g. `clear-js-timeout!` does not need a docstring that says "Clears a js timeout"...it adds no clarity).

If you want to add doc strings, then please consider the programmer that will be using the function, and what they are going to want to know:

- What are the arguments (i.e. allowed types, units, etc.).  If an argument is named `tm` for example, is that an `inst?`, and integer? Is it measured in seconds, minutes, milliseconds? 
- What is the return value
- Are there other functions that they should consider/look at?

Also, try to format the docstring using markdown, since cljdoc and other tools support that. For example:

```
(defn set-js-timeout!
  "Create a timer that will call `f` after `tm` milliseconds.

   Returns an opaque value that can be passed to `clear-js-timeout!` to cancel the timer.

  See also `defer`, which is intended to be platform independent.
  "
  [tm f]
  ...)
```

abhi18av:
Thanks @awkay , exactly the guidlines I was looking for. Maybe, we should really be adding these bits to the [contributing.md](https://github.com/fulcrologic/fulcro/blob/develop/CONTRIBUTING.md) which would make this knowledge referenceable :)

awkay:
If you want to do that, please feel free.

abhi18av:
Dropped some docstrings which I couldn't flesh out with confidence.

PR #356, opened 2020-01-15
upgrade dom/wrap-form-element to use componentDidUpdate; handle deleted props;

thosmos:
I made a fix to handle the problem of props disappearing and not getting undefined from React local state.  In the process I upgraded it to use componentDidUpdate instead of the deprecated UNSAFE_componentWillReceiveProps.  I haven't tested it yet beyond the user's test case on Slack.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/dom.cljs
+++ b/src/main/com/fulcrologic/fulcro/dom.cljs
@@ -155,23 +155,50 @@
           (update-state
             this (.-props this)
             (gobj/getValueByKeys event "target" "value"))))
+      (componentDidUpdate [this prev-props prev-state snapshot]
+        (let [state-value    (gobj/getValueByKeys this "state" "value")
+              this-node      (js/ReactDOM.findDOMNode this)
+              value-node     (if (is-form-element? this-node)
+                               this-node
+                               (gdom/findNode this-node #(is-form-element? %)))
+              element-value  (gobj/get value-node "value")
 
-      (UNSAFE_componentWillReceiveProps [this new-props]
-        (let [state-value   (gobj/getValueByKeys this "state" "value")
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/dom.cljs
```diff
@@ -155,23 +155,50 @@
...
+                               props)]
+
+          ;; delete keys for props that went away
+          ;; ... setState is always a merge and the only way to "delete" keys is to set to nil
+          (when props-changed?
+            (doseq [old-k (gobj/getKeys prev-props)]
+              (when-not (gobj/get next-props old-k)
+                (gobj/set next-props old-k nil))))
 
-      (UNSAFE_componentWillReceiveProps [this new-props]
```
awkay: Nope...you have to use the UNSAFE variant or you'll get React warnings (and even crashes) in newer versions...this one is going away.

awkay:
So, this isn't going to work with React in later versions per comment I've added.  ALSO, this is a very sensitive area of code (possibly one of the most sensitive in the code base)...all sorts of input field misbehaviors happen if you get this wrong.  You need to test things like:

- does the cursor jump when editing the middle of a text field when controlled from fulcro state
- Does the value get out of sync when you change props and refresh locally (refresh-only)?
- Does it work right if you control the value with component local state?

etc.

I hate messing with this bit of code because it is so easy to break lots of apps in weird and hard-to-detect ways.

thosmos:
there are [better](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props) [ways](https://github.com/codonnell/fulcro/commit/2542259d5fce6d6ca8221e6f869c8e5d9580d657) to  fix these problems

Issue #357, opened 2020-01-15
dom/wrap-form-element fails to remove missing props from state


thosmos:
When a component adds and then removes a key to props, something like:
```clojure
     (ui-form-input (cond-> {:onChange (fn [evt]
                                         (m/set-string! this :user/name :event evt)
                                         (comp/transact! this [(fs/mark-complete! {:field :user/name})]))
                             :value name}
                      (fs/invalid-spec? user :user/name)
                      (assoc :error "User name cannot be blank")))
```
then the removed key does not also get removed from state.  

@codonnell has a fix here: https://github.com/codonnell/fulcro/commit/2542259d5fce6d6ca8221e6f869c8e5d9580d657
which moves the component's state under one child key called `data`.  This enables to component to add/remove keys, but may have consequences if existing users of this wrapper expect those values to be in the root of the state variable.  Is this possible?

thosmos:
oops

thosmos:
fixed in 9b2828131a0e4d5c098c0801542ea0b1d192cfd5

PR #358, opened 2020-01-15
fixes dom/wrap-form-element to remove disappearing props from state

thosmos:
This is as minimal of a fix as I could come up with for #357.  It only touches state when there are deleted keys by adding nils.


thosmos:
need to make a change and squash

PR #359, opened 2020-01-15
fixes dom/wrap-form-element to remove disappearing props from state

thosmos:
This is as minimal of a fix as I could come up with for #357. It only touches state when there are deleted keys by adding nils.


thosmos:
fixed in 9b2828131a0e4d5c098c0801542ea0b1d192cfd5

PR #363, opened 2020-01-23
add ion-middleware namespace which uses a version of Transit compatib…

mdhaney:
…le with Datomic Cloud Ions

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/transit_old.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/transit_old.cljc
@@ -1,0 +1,109 @@
+(ns com.fulcrologic.fulcro.algorithms.transit-old
+  "This version is for use with older versions of Transit, such as what is supported on Datomic Cloud
+
+   Transit functions for the on-the-wire EDN communication to common remotes. Includes support for Fulcro tempids,
+   and can be extended to support additional application-specific data types."
+  #?(:clj
+     (:refer-clojure :exclude [ref]))
+  (:require [cognitect.transit :as t]
+            #?(:cljs [com.cognitect.transit :as ct])
+            [com.fulcrologic.fulcro.algorithms.tempid :as tempid #?@(:cljs [:refer [TempId]])])
+  #?(:clj
+     (:import [com.cognitect.transit
+               TransitFactory WriteHandler ReadHandler]
+              [com.fulcrologic.fulcro.algorithms.tempid TempId])))
+
+#?(:cljs
+   (deftype TempIdHandler []
+            Object
+            (tag [_ _] tempid/tag)
+            (rep [_ r] (. r -id))
+            (stringRep [_ _] nil)))
+
+#?(:clj
+   (deftype TempIdHandler []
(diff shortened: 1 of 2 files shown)


src/main/com/fulcrologic/fulcro/algorithms/transit_old.cljc
```diff
@@ -0,0 +1,109 @@
...
+     (:refer-clojure :exclude [ref]))
+  (:require [cognitect.transit :as t]
+            #?(:cljs [com.cognitect.transit :as ct])
+            [com.fulcrologic.fulcro.algorithms.tempid :as tempid #?@(:cljs [:refer [TempId]])])
+  #?(:clj
+     (:import [com.cognitect.transit
+               TransitFactory WriteHandler ReadHandler]
+              [com.fulcrologic.fulcro.algorithms.tempid TempId])))
+
+#?(:cljs
```
awkay: Is it really necessary to have definitions of these in separate namespaces?  Perhaps it would make sense to centralize these somehow?  Or was it that you would need to move them, which would be an incompatible change?

src/main/com/fulcrologic/fulcro/algorithms/transit_old.cljc
```diff
@@ -0,0 +1,109 @@
...
+  #?(:clj  (try
+             (.write (writer (java.io.ByteArrayOutputStream.)) v)
+             true
+             (catch Exception e false))
+     :cljs (try
+             (.write (writer) v)
+             true
+             (catch :default e false))))
+
+(defn transit-clj->str
```
awkay: Do we need these?  If you're doing SSR you really HAVE to use metadata, so these just won't work for their intended purpose.

src/main/com/fulcrologic/fulcro/server/ion_middleware.clj
```diff
@@ -0,0 +1,182 @@
...
+  (:require
+    [clojure.pprint :refer [pprint]]
+    [clojure.test :refer :all]
+    [clojure.repl :refer [doc source]]
+    [com.fulcrologic.fulcro.algorithms.transit-old :as transit]
+    [cognitect.transit :as ct]
+    [taoensso.timbre :as log])
+  (:import (java.io ByteArrayOutputStream)))
+
+(def not-found-handler
```
awkay: Quite a few of these functions are purely duplicate code. Is this necessary?

mdhaney: Yes, I hate all the duplication, too.  I’ve thought of a way to reduce the amount of duplication.  I’ll make the changes and update the PR, probably tomorrow.

awkay:
Any progress on this?

mdhaney:
Sorry, I’ve been putting out fires all week.  It will probably be another week or two before I can realistically spend time on this, so if you want to close this PR for now, that’s fine.  I’ll take another stab at it when things settle down a bit.

awkay:
Yeah thanks.  When you get back to it feel free to open up a new PR.

PR #364, opened 2020-01-27
added optional xhrio param to c.f.f.n.http-remote/fulcro-http-remote

fjolne:
This allows to pass customized XhrIo instance to fulcro-http-remote, while preserving default behaviour.

Common case for customization is React Native, where all the requests are done cross-origin, and XhrIo requires withCredentials flag in order to set cookies from the server response. https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials


fjolne:
NVM, need to pass `make-xhrio` itself as `xhrio` should be created per request. Closing PR.

PR #366, opened 2020-01-28
added optional channel-type to c.f.f.i.inspect-ws/start-ws-messaging

fjolne:
Added optional `channel-type` param, which would allow to remove error when connecting to Inspect Electron from React Native.


awkay:
Thanks for your work on this!

PR #369, opened 2020-02-03
use key into component registry instead of component class

davewo:
maintain get- and set- query api

### DIFF
--- a/src/main/com/fulcrologic/fulcro/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/components.cljc
@@ -522,7 +522,11 @@
 (defn denormalize-query
   "Takes a state map that may contain normalized queries and a query ID. Returns the stored query or nil."
   [state-map ID]
-  (let [get-stored-query (fn [id] (get-in state-map [::queries id :query]))]
+  (let [get-stored-query (fn [id]
+                           (let [{:keys [query component-key]} (get-in state-map [::queries id])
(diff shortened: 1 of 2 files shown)


awkay:
Did you find the automated tests for this in the code???

davewo:
yep I found them, I will keep this PR in draft as I rework the change based on our conversation

src/main/com/fulcrologic/fulcro/components.cljc
```diff
@@ -799,13 +805,16 @@
...
-  (let [new-state (normalize-query-elements state-map query)
-        new-state (if (nil? (::queries new-state))
-                    (assoc new-state ::queries {})
-                    new-state)
-        top-query (link-query query)]
-    (if-let [queryid (some-> query meta :queryid)]
-      (util/deep-merge {::queries {queryid {:query top-query :id queryid}}} new-state)
+  (let [queryid       (some-> query meta :queryid)
+        component-key (class->registry-key (some-> query meta :component))
+        query'        (vary-meta query dissoc :queryid :component)
```
awkay: Why is this necessary? (clearing the metadata)...oh, is this so it does not end up in state map?

davewo: That was my thought, keeping metadata out of state. If you think its OK to leave it then I will update...

PR #370, opened 2020-02-10
fix pristine->entity* to remove keys that got added after pristine

thosmos:
This fixes the problem where a form adds keys that were missing when pristine was created, but then attempts to reset.  Currently the reset will fail because `pristine->entity*` only merges pristine back into the entity and leaves the new keys, which leaves the form in a dirty state.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
@@ -550,8 +550,13 @@
   Only affects declared fields and sub-forms."
   [state-map entity-ident]
   [map? eql/ident? => map?]
-  (update-forms state-map (fn reset-form-step [e {:keys [::pristine-state] :as config}]
-                            [(merge e pristine-state) config]) entity-ident))
+  (update-forms state-map
+    (fn reset-form-step [e {:keys [::pristine-state ::fields] :as config}]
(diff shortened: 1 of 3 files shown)


awkay:
Thanks for working on this long-standing annoyance :)

Tests?

thosmos:
Oh yeah, tests!  I'm working on a deadline and would love to revisit this later this week or weekend to make some tests.

thosmos:
OK I added a test that fails without the fix

src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
```diff
@@ -550,8 +551,9 @@
   Only affects declared fields and sub-forms."
   [state-map entity-ident]
   [map? eql/ident? => map?]
-  (update-forms state-map (fn reset-form-step [e {:keys [::pristine-state] :as config}]
-                            [(merge e pristine-state) config]) entity-ident))
```
awkay: Why not just:

```
(as-> e e (apply dissoc e (::fields config)) (merge e pristine-state))
```

thosmos: yes that's better :)

awkay:
OK, I did the simpler fix, and beefed up the test in both directions. Released in 3.1.9.

PR #377, opened 2020-03-02
Warn if route ident differs from component's

holyjak:
If the ident passed to dr/route-immediate / -deferred differs from
the ident of the component, log a warning.

Example log (logged 4 times :-( ):
> backend.js:6 WARN [com.fulcrologic.fulcro.routing.dynamic-routing:103] -  route-immediate  was invoked with the ident  [:offboarding/id "DUMMYID"]  which doesn't seem to match the ident of the wrapping component ([ :component/id  ...])

### DIFF
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -103,7 +103,7 @@
   (when (and #?(:clj false :cljs goog.DEBUG)
              *target-class*
              (not (ident-matches-expectation? (comp/ident *target-class* {}) ident)))
-    (log/warn fn-name " was invoked with the ident " ident
+    (log/error fn-name " was invoked with the ident " ident
               " which doesn't seem to match the ident of the wrapping component (class "
               *target-class* " , ident ["
               (first (comp/ident *target-class* {})) " ...])")))
@@ -154,6 +154,9 @@
       (if router-id
         (do
           (log/debug "Router" router-id "notified that pending route is ready.")
+          (when (and #?(:clj false :cljs goog.DEBUG) (nil? (get-in state-map target)))
+            (log/error `target-ready "should route to" target "but there is no data in the DB for the ident."
+                      "Perhaps you supplied a wrong ident?"))
           (uism/trigger! app router-id :ready!))
         (log/error "dr/target-ready! was called but there was no router waiting for the target listed: " target
           "This could mean you sent one ident, and indicated ready on another."))))
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -22,6 +22,10 @@
     #?(:clj [cljs.analyzer :as ana])
     [com.fulcrologic.fulcro.algorithms.indexing :as indexing]))
 
+;; Class of the routing target component, available in the notifications fns
+;; (:will-enter, :route-cancelled, :will-leave)
+(def ^:dynamic *target-class* nil)
+
 (declare route-immediate)
 
 (defn route-segment
@@ -39,7 +43,8 @@
   "Universal CLJC version of will-enter.  Don't use the protocol method in CLJ."
   [class route-params]
   (when-let [f (get-route-cancelled class)]
-    (f route-params)))
+    (binding [*target-class* class]
+      (f route-params))))
 
 (defn get-will-enter
   "Returns the function that is called before a route target is activated (if the route segment of interest has changed and the
@@ -62,7 +67,8 @@
   "Universal CLJC version of will-enter."
   [class app params]
   (when-let [will-enter (get-will-enter class)]
-    (will-enter app params)))
+    (binding [*target-class* class]
+      (will-enter app params))))
 
 (defn route-target? [component] (boolean (comp/component-options component :route-segment)))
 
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -22,6 +22,11 @@
     #?(:clj [cljs.analyzer :as ana])
     [com.fulcrologic.fulcro.algorithms.indexing :as indexing]))
 
+;; ident of the current component, derived from its class.
+;; Since we don't know its props yet, if the "id" part depends on props,
+;; it will be `nil`.
+(def ^:dynamic *current-comp-ident* nil)
```
awkay: I'm not loving the dynamic var. We're exposing something for convenience here that isn't terribly useful for anything but this error check. It might make sense to instead expose something that's a bit more useful, such as the class itself? The thing is I don't want something globally visible like this that isn't generally useful, and since we also don't have a clear (documented) story of when that will be valid it could be quite confusing, since it is only valid in very special circumstances.

holyjak: Ok, I have changed that to `*target-class* and set it now for each of the three notification functions. Is that better? If not, do you have any tip how to solve it differently? How else could route-immediate/... get the info inside which component it is invoked?

holyjak:
Question: Is there any simple way to log the warning only once and not every time we run `:will-enter` during the same route resolution? I suppose not and it is better to log the warning 4 times in a row then at all... :-)

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -86,9 +93,29 @@
...
+  ;; NOTE: If the `id` of the ident is hardcoded then maybe-expected-id will be set,
+  ;; but if it depends on props then it will be nil
+  (and (= expected-table table)
+       (or (nil? maybe-expected-id) (= maybe-expected-id id))))
+
+(defn- check-ident-matches-expectation? [fn-name ident]
+  (when (and #?(:clj false :cljs goog.DEBUG)
+             *target-class*
+             (not (ident-matches-expectation? (comp/ident *target-class* {}) ident)))
+    (log/warn fn-name " was invoked with the ident " ident
```
holyjak: should this be a warning or rather an error? Perhaps error?

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -127,6 +154,9 @@
       (if router-id
         (do
           (log/debug "Router" router-id "notified that pending route is ready.")
+          (when (and #?(:clj false :cljs goog.DEBUG) (nil? (get-in state-map target)))
+            (log/warn `target-ready "should route to" target "but there is no data in the DB for the ident."
```
holyjak: 1. warn of error?
2. Can we think of an even better error message?

awkay: seems ok.

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
awkay: yeah, error. You're already allowing for the possible nil id, so the only reason you would be logging is if it really looks like an error.

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
awkay: Good enough for me.

holyjak:
I have changed both log/warn to errors. Are we good to merge?

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -22,6 +22,10 @@
     #?(:clj [cljs.analyzer :as ana])
     [com.fulcrologic.fulcro.algorithms.indexing :as indexing]))
 
+;; Class of the routing target component, available in the notifications fns
+;; (:will-enter, :route-cancelled, :will-leave)
```
awkay: Convert this comment to a docstring pls.

awkay: Also, why is this available in route-cancelled and will-leave?  Are you just saying that because it's true, or because you see a reason?  I don't like to make API guarantees that we don't have to.

holyjak: Yes, it is true. I mentioned previously you did not like the dynamic var also because it was only for that one use case. So I extended it to all the notification functions, see L47 and L86.

I will convert the comment to a docstring but the var is not intended for external use, I think. Would it then make sense to annotate it with [`^:no-doc`](https://github.com/cljdoc/cljdoc/blob/master/doc/userguide/for-library-authors.adoc#hiding-namespaces--vars-in-documentation)?

awkay: We could go 2 ways. If it is only ever used by warnings, then perhaps it isn't a documented thing, and instead should be surrounded by `when goog.DEBUG` so it is elided completely in production, making it obvious it is just an add-on for warnings.

holyjak:
What do you prefer?

awkay:
Good enough...I'll touch it up :)

holyjak:
❤️

PR #383, opened 2020-03-08
Error if get-query in :query outside of a join

holyjak:
E.g. `:query [:my/key (comp/get-query MyChild)]` will fail with

> defsc MyComponent: `get-query` calls in :query can only be inside a join value,
> i.e. `{:some/key (comp/get-query Header)}` at ...

Note: I do not include the preceeding key so that I also catch
`:query [(get-query AnotherChild)]` and b/c I cannot know whether
it is intended as a key for the join or an unrelated property.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/components.cljc
@@ -1181,9 +1181,13 @@
                queried-keywords      (-legal-keys template)
                has-wildcard?         (some #{'*} template)
                to-sym                (fn [k] (symbol (namespace k) (name k)))
-               illegal-syms          (mapv to-sym (set/difference destructured-keywords queried-keywords))]
+               illegal-syms          (mapv to-sym (set/difference destructured-keywords queried-keywords))
+               component-query       #(and (list? %) (= "get-query" (name (first %))) %)
+               err-env               (merge env (meta template))]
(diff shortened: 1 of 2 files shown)


holyjak:
Should I perhaps add a test for this case to `test/com/fulcrologic/fulcro/macros/defsc_spec.clj` / somewhere?

src/main/com/fulcrologic/fulcro/components.cljc
```diff
@@ -1181,9 +1181,13 @@
                queried-keywords      (-legal-keys template)
                has-wildcard?         (some #{'*} template)
                to-sym                (fn [k] (symbol (namespace k) (name k)))
-               illegal-syms          (mapv to-sym (set/difference destructured-keywords queried-keywords))]
+               illegal-syms          (mapv to-sym (set/difference destructured-keywords queried-keywords))
+               component-query       #(and (list? %) (= (name `get-query) (name (first %))) %)
```
awkay: `(= 'query (name (first %)))` perhaps?

holyjak: I guess you meant replacing 
```
(name `get-query)
```
with `"get-query"`? I prefer to use the actual symbol instead of a hardcoded string so that if you ever rename it, it wouldn't break. But given how that is unlikely string is fine, if that is what you prefer.

PR #392, opened 2020-04-02
Major Interesting Improvements:

awkay:
- Added `transact!` open `:synchronous?` which allows for use of raw inputs without cursor jumping, and with much faster performance.
- Added `transact!!` as a shorthand for turning on sync option in `transact!`
- Added compiler option `:wrap-inputs?` to switch DOM to generate raw inputs. Defaults to true for bw compat.
- Added compiler option `:default-to-hooks?` to enable hooks by default on `defsc`.
- Changed mutation helpers to use `transact!!`
- Formalized props tunneling API into `comp/tunnel-props!`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>com.fulcrologic</groupId>
   <artifactId>fulcro</artifactId>
   <packaging>jar</packaging>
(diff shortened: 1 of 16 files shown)


wilkerlucio:
Awesom!!

PR #396, opened 2020-04-09
Fix children on hook components

wilkerlucio:
Before this patch, `prim/children` always returns `nil` on the hook components.


awkay:
Deployed in 3.2.1-SNAPSHOT

PR #403, opened 2020-04-11
Always use http to connect in fulcro inspect websockets

wilkerlucio:
So in electron, in which the URL is `file://`, this makes it work there.


awkay:
Can't do it.  Browsers won't allow mixed content on that. Perhaps if you do detection that it isn't a restricted environment that would be ok, but not sure how you'd do that reliably.

wilkerlucio:
But in case of `https`, does it work? Does inspect have a `https` version of WS running as well? If not would make no difference, but I'm ok adding the check anyway, I can do it in the same way Sente does it.

PR #406, opened 2020-04-29
fixed passing down dr/will-enter timeouts-and-params args

fjolne:
When calling `dr/change-route!` or `dr/change-route-relative!` according to documentation it should be possible to pass extra params to `:will-enter` of the target component via the last `timeouts-and-params` arg.

But as `:will-enter` is called several times during the path resolution, some of this calls (particularly made by `proposed-new-path` and (transitively) by `signal-router-leaving`) lose this extra params, and `:will-enter` is called only with `:route-segment` params.

That makes it difficult to do deferred routing on non-singleton components. Consider the case of "tabs", when there's a root "tab manager" component `Company` and child "tabs" which are responsible for their data load, but need initial data from the parent component to do the load (`company/id` in the case):
```clojure
(defsc CompanyVacancies [this {:company/keys [vacancies]}]
  {:query         [:company/id
                   {:company/vacancies (get-query Vacancy)}]
   :ident         :company/id
   :route-segment ["vacancies"]
   :will-enter    (fn [app {:company/keys [id]}]
                    (let [id (uuid id)]
                      (dr/route-deferred [:company/id id]
                        #(df/load! app [:company/id id] CompanyVacancies
                           {:post-mutation        `dr/target-ready
                            :post-mutation-params {:target [:company/id id]}}))))})

(defrouter CompanyRouter [this props]
  {:router-targets [vacancies/CompanyVacancies ,,,]})
(def ui-company-router (comp/factory CompanyRouter))

(defsc Company [this {:ui/keys [router]}]
  {:query         [{:ui/router (get-query CompanyRouter)}]
   :ident         (fn [] [:component/id :company])
   :initial-state {:ui/router {}}
   :route-segment ["company" :company/id]
   :will-enter    (fn [app {:company/keys [id]}]
                    (let [id (uuid id)]
                      (dr/route-deferred [:component/id :company]
                        #(do
                           (dr/change-route-relative! app Company ["vacancies"]
                             {:company/id (str id)})
                           (df/load! app :some-data nil
                             {:post-mutation        `dr/target-ready
                              :post-mutation-params {:target [:component/id :company]}})))))}
  (ui-company-router router))
```

This PR passes `timeouts-and-params` arg down to `proposed-new-path` and `signal-router-leaving`, leaving the signatures of this functions backward-compatible (empty map if no `timeouts-and-params` are passed).

### DIFF
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -386,7 +386,9 @@
 (defn proposed-new-path
   "Internal algorithm: Returns a sequence of idents of the targets that the `new-route` goes through by analyzing the current
   application query and state."
-  [this-or-app relative-class-or-instance new-route]
+  ([this-or-app relative-class-or-instance new-route]
+   (proposed-new-path this-or-app relative-class-or-instance new-route {}))
+  ([this-or-app relative-class-or-instance new-route timeouts-and-params]
   (let [app        (comp/any->app this-or-app)
         state-map  (app/current-state app)
         router     relative-class-or-instance
@@ -403,7 +405,7 @@
               segment        (route-segment target)
               params         (reduce
                                (fn [p [k v]] (if (keyword? k) (assoc p k v) p))
-                               {}
+                                (dissoc timeouts-and-params :error-timeout :deferred-timeout)
                                (map (fn [a b] [a b]) segment matching-prefix))
               target-ident   (will-enter target app params)]
           (when (or (not (eql/ident? target-ident)) (nil? (second target-ident)))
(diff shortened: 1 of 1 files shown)


awkay:
Great, would you mind terribly fixing this to not submit whitespace changes?  Reformatting code  you didn't touch makes history unclear.

putting the following in your global .gitconfig on OSX or Linux:

```
[alias]
  addnw=!sh -c 'git diff -U0 -w --no-color "$@" | git apply --cached --ignore-whitespace --unidiff-zero -'
```

will give you a `git addnw .` that only adds real changes, allowing you to commit your changes without formatting changes.  

so if you `git reset HEAD^` on your branch, then `git addnw .` and re-commit with a force push it should fix it.

fjolne:
Wow, thanks a lot for the tip, it's awesome! I thought of making it as `[... & [timeouts-and-params]]` to have less whitespace interference, but that would be inconsistent with your code style and nil punning comes to play... Good to know there's a better way!

awkay:
on 3.2.4-SNAPSHOT

PR #407, opened 2020-05-05
Use only-refresh when triggering toggle! or set-value!

wilkerlucio:
While testing performance under large collections, this simple change has a significant impact.

Running without `:only-refresh`

v25nfS-XVQ

Running with `:only-refresh`

GlLZCALGlT


awkay:
I fixed this a different way. There are not `!!` versions of these that use the props tunneling. Should be even faster, and won't break bw compat.

Issue #409, opened 2020-05-18
Query transforms do not preserve pre-merge semantics


RokLenarcic:
Here I am talking about code shown in the Developer Manual http://book.fulcrologic.com/#_pre_merge_basics

In this example the key `:ui/count` is initialized by pre-merge hook to be `{:ui/count 5}`.

It is not mentioned in the Manual that this only works incidentally, because `:ui/count` is a simple value that requires no normalization. This same approach doesn't work for `fs/form-config-join` for instance.

To be more concrete:

Let's take the component from the manual and add form state to it. The resulting component query is:

`[::counter-id ::counter-label :ui/count fs/form-config-join]`

When issuing a load to the server, the UI centric data get's elided from the query by the default (which is what we want), so the server gets query:

`[::counter-id ::counter-label]`

Server returns some data and our pre-merge hook is called. 

And here's the issue: any data our pre-merge hook returns will be normalized with the new, after-elision, query. So whatever we return besides `::counter-id` and `::counter-label` from the pre-merge hook will be merged into state as-is.

In the example in the manual, it makes no difference to `:ui/count 5` whether normalization works or not. But when you try to add form config to incoming data:

```
(merge current-normalized
       (fs/add-form-config Countdown data-tree))
```

it breaks, throwing errors because form state gets added unnormalized. 

To me this seems like a broken feature, because it only works for the simple case, and it is not clear (nor explained in the docs) why adding form config in pre-merge doesn't work (it requires quite a bit of debugging to figure it out). So even if this is intended behaviour it lack clarity.

It feels to me that adding UI centric data to incoming data in a component focused, component local fashion is what pre-merge was made for, but normalizing only properties that were queried to server kills the whole thing. The workaround of turning off eliding of UI centric properties isn't too hot either.

awkay:
@wilkerlucio do you remember the aspects of this with respect to query transform?  I thought we were using the pre-elision query here, but perhaps it got changed for some other competing reason?  I'll look at the current logic.

awkay:
So, @RokLenarcic I agree with you that this isn't ideal.  There are three competing features colliding in this particular code location:

* You can transform the query in `load!` itself.
* You can use the global transform to change the query (elide nodes).
* The remote code itself can change the query to match the data (advanced feature that might not be the best idea now...was originally added for file upload progress, but could also be used to merge alternate error data into the db).

[...]

PR #410, opened 2020-05-18
Load merge fix

awkay:
Changes load completion to use the proper query for actual merge.


awkay:
This should fix issue #409 , but I wonder if some other eyes could look at the diff and make sure no red flags pop up.

awkay:
Pushed to clojars as 3.2.7-SNAPSHOT

awkay:
This can technically break existing applications, but I think the risk is quite low, and I can easily argue that morphing the query for merge from a load at layers underneath load should really not be done.  That said, there is a workaround if you want to do it: provide a custom internal-load mutation, so even the old behavior can still be had.

Issue #412, opened 2020-05-25
default-global-eql-transform elides i18n query


fjolne:
`fulcro-i18n` uses `::i18n/translations` server property to load translations, but it is elided by default, and it seems like there's no way to make it work other than to provide a custom `:global-eql-transform`.

Maybe we could either:
- make `app/default-global-eql-transform` more customizable (there's no way to plug in a predicate into it right now, only fully rewrite it)
- provide some way to bypass / modify `global-eql-transform` from `df/load!`
- change `default-global-eql-transform` to elide `(str/starts-with? ns "com.fulcrologic.fulcro.")` (notice the dot in the end, which would elide only core Fulcro namespaces).

awkay:
Oops. I really have not had time to work on the i18n stuff, and missed that regression. It should certainly not elide that one, but there are dynamic routing, form state, and other core library ones that it should. I think at the moment probably just changing it to not elide that one in particular is what I'd prefer. Basically any library feature of Fulcro that does not need server-side data should be elided, so I'd rather add exceptions.  I'm not up to generalizing it any more. The tools to make your own are present, and it is easy enough to just make your own function.

fjolne:
This PR isn't particularly what you asked for (making an exception for i18n), but it seems to do what you want: it elides routing, form state, etc.

I'd think that it's better to elide less than more because
- it makes `default-global-eql-transform` composable with custom transforms
- it makes future Fulcrologic libs in charge of their queries elision (via `load!` `:update-query`) => no need to make an exception in the core Fulcro repo every time new Fulcrologic lib needs it

PR #413, opened 2020-05-26
elide only core Fulcro namespaces from remote query

fjolne:
https://github.com/fulcrologic/fulcro/issues/412


awkay:
ok, fine by me

PR #421, opened 2020-06-16
file-upload: add response-type to outgoing request

strssndktn:
This patch fixes a bug which let's the default response middleware not
extract the response originated by a file-upload originated request.

While introducing support for mutations returning binary data in commit
0ceb296cb34a380, a check was introduced to match up the response-type
of the request with that of the incoming response. Because file-upload
doesn't set response-type, the transit decoding was bypassed.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/networking/file_upload.cljs
+++ b/src/main/com/fulcrologic/fulcro/networking/file_upload.cljs
@@ -5,6 +5,7 @@
     [edn-query-language.core :as eql]
     [com.fulcrologic.guardrails.core :refer [>defn => >def]]
     [com.fulcrologic.fulcro.algorithms.transit :as t]
+    [com.fulcrologic.fulcro.networking.http-remote :as http]
     [taoensso.timbre :as log]))
 
 (defn new-upload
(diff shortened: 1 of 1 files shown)


awkay:
I'm a little confused.  I'm actively using this support and it works for me. Could you be a bit more detailed in your description of what goes wrong?  File upload doesn't set response type, and you say transit decoding is bypassed...what is the end result of that?  What problem do you see?    I appreciate that you took the time to figure this out and trace it down, but since a lot of code depends on this I feel a little odd merging it in without at least **one** of the following:

1. Manual tests against some existing app that uses it, like the fulcro-rad-demo, that show it doesn't break existing users.
2. A demo of what goes wrong (repro case) that is fixed by it
3. A clearer explanation (with pointers to line numbers in source) of what you discovered. (when I do X, line Y does Z, and that causes R).

strssndktn:
The problem appeared when I tried to return data from the pathom handler that processed the file upload, in this particular case the :tempids mapping. Without returning data from the mutatoris that handle file uploads, no breakage is visible.

While debugging the code why my tempids won't get translated I noticed that in the response middleware the body was merely a string and thus all lookups to :tempids didn't resolve, thus leaving my tempids untranslated.

I checked if this could have affected any demo projects but e.g. the fulcro-rad demo doesn't use any "return-data" from the blob upload subsystem.

In particular, while processing the response from an upload, the `ct/read` doesn't get executed in this code:

```clojure
     (fn fulcro-response-handler [{:keys [body error outgoing-request] :as response}]
       (handler
         (let [{:keys [response-type]} outgoing-request]
           (if (= :default response-type)
             (try
               (if (= :network-error error)
                 response
                 (let [new-body (if (str/blank? body)
                                  {}
                                  (ct/read reader body))
                       response (assoc response :body new-body)]
                   response))

```

awkay:
Thanks! Good work, and thanks for the details.

PR #426, opened 2020-09-11
Rename ::mutations/mutation-error to :ui/...

holyjak:
so that interested components can easily request it in their queries and it is not sent to the server.

This is a possibly breaking change for users that already use the key `:ui/mutation-error` in their code or rely on the old name. Alternatives to this solution:

* Use a different key where the likelihood of a conflict is smaller, e.g. `:ui/fulcro-mutation-error`. But it isn't that nice...
* Keep the name as is but add it to the default blacklist so that it is not sent to the server.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/mutations.cljc
+++ b/src/main/com/fulcrologic/fulcro/mutations.cljc
@@ -78,14 +78,14 @@
 
 (>defn update-errors-on-ui-component!
   "A handler for mutation network results that will place an error, if detected in env, on the data at `ref`.
-  Errors are placed at `k` (defaults to `::m/mutation-error`).
+  Errors are placed at `k` (defaults to `:ui/mutation-error`).
 
   Typically used as part of the construction of a global default result handler for mutations.
 
   Swaps against app state and returns `env`."
   ([env]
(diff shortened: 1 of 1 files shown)


awkay:
I agree with your intent but not your implementation. Breaking is not acceptable. Add it to `:config` in app (which is env) and then you can configure the key name on app startup.

holyjak:
I thought so :-) I will look into it.

holyjak:
After thinking some more, making the keyword configurable would be nice -
but how much value does it really provide? The current name is just fine
and I don't really see why would somebody need to change it. The only
current problem is that it will be included in the query sent to Pathom.
Wouldn't a better solution, which would provide value out of the box for
everybody, be adding it to the default blacklist next to :ui/*?

awkay:
I agree that the blacklist solution might just be best.

awkay:
Are you sure this is an issue at all? I'm looking at the built-in stuff and it does elide anything whose ns starts with com.fulcrologic.fulcro, which mutation-errors clearly does. Have you overridden the eql transform?

I'm closing this for now, as I don't actually see a problem with the built-in code.

PR #428, opened 2020-09-23
Fix in HTTP remote

currentoor:
`eql/ast->query` already returns a valid EQL query, which is a vector. For some reason it was being wrapped by an extra vector. There was a breaking change in EQL version 0.0.10.

https://github.com/edn-query-language/eql/blob/master/CHANGELOG.md#0010

Note this is consistent with what's in the file-upload namespace.

https://github.com/fulcrologic/fulcro/blob/develop/src/main/com/fulcrologic/fulcro/networking/file_upload.cljs#L98


awkay:
Yes, that was an issue. I actually had a correct workaround in my do-no-use namespace, but was not using it ;p

I had a couple more things to change, so I did this by hand. Thanks for looking into it.

PR #434, opened 2020-10-30
Optimize comp/fragment

nivekuil:
1. The apply seems extraneous

2. Avoid calling clj->js when no props are given, seems to save ~.2us for me.

3. The & binding form is faster:

cljs.user> (simple-benchmark [seq (range 50)] (let [[x y] [(first seq) (rest seq)]]) 1000000)
[seq (range 50)], (let [[x y] [(first seq) (rest seq)]]), 1000000 runs, 116 msecs
cljs.user> (simple-benchmark [seq (range 50)] (let [[x y] [(first seq) (rest seq)]]) 1000000)
[seq (range 50)], (let [[x y] [(first seq) (rest seq)]]), 1000000 runs, 117 msecs
cljs.user> (simple-benchmark [seq (range 50)] (let [[x y] [(first seq) (rest seq)]]) 1000000)
[seq (range 50)], (let [[x y] [(first seq) (rest seq)]]), 1000000 runs, 122 msecs

cljs.user> (simple-benchmark [seq (range 50)] (let [[x & y] seq]) 1000000)
[seq (range 50)], (let [[x & y] seq]), 1000000 runs, 59 msecs
cljs.user> (simple-benchmark [seq (range 50)] (let [[x & y] seq]) 1000000)
[seq (range 50)], (let [[x & y] seq]), 1000000 runs, 52 msecs
cljs.user> (simple-benchmark [seq (range 50)] (let [[x & y] seq]) 1000000)
[seq (range 50)], (let [[x & y] seq]), 1000000 runs, 54 msecs

### DIFF
--- a/src/main/com/fulcrologic/fulcro/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/components.cljc
@@ -1214,10 +1214,10 @@
                           [{} args])]
        (vec children))
      :cljs
-     (let [[props children] (if (map? (first args))
-                              [(first args) (rest args)]
-                              [#js {} args])]
-       (apply js/React.createElement js/React.Fragment (clj->js props) (force-children children)))))
+     (let [[props & children] args]
+       (if (map? props)
+         (js/React.createElement js/React.Fragment (clj->js props) (force-children children))
+         (js/React.createElement js/React.Fragment #js {} (force-children args))))))
 
 #?(:clj
    (defmacro with-parent-context


awkay:
Curios why you felt the need to optimize this?

awkay:
I tend not to accept these for a number of reasons:

* I don't remember under what circumstances the `apply` mattered to me...react version perhaps?
* The code is working as-is
* I don't have time to regression test things
* I'm not confident that the contributor regression tested to my standards (i.e. multiple react versions, did you try when they supply a lazy seq, a vector, hand-written children, etc.)

The performance value here has not been a complaint from anyone, and the possibility of *breaking* someone is non-zero. Thus, I'm not terribly interested. I appreciate that you want to help, but this kind of optimization belongs in your own code. I'd accept it if there was a known problem from a known use-case...but just asking to do a little opt like this because it "looks slower than necessary" is not good enough.

nivekuil:
Got it -- stability is good.

On 30 October 2020 15:55, Tony Kay <notifications@github.com> wrote:

> Curios why you felt the need to optimize this?

I use a lot of fragments, at least a few hundred per render, and I think
I rewrote this one when I was first looking at what fragment actually
did.  At this point I have a few changes to Fulcro locally; don't want
to get too far off of upstream.

PR #437, opened 2020-12-10
Fix for element picker on React 17

rvllo:
Fulcro Inspect's element picker wasn't working with a clean install of the latest fulcro-template. Neither with the chrome extension nor with the electron inspector app.

After investigating, I found the problem to be that react 17 changed the components internal instance property.


awkay:
Great, thanks for looking into it!

PR #438, opened 2020-12-14
defrouter: clarify limitations of default target

holyjak:
I originally believed that relying on the router to display the default target is the same as routing to it explicitly.  https://book.fulcrologic.com/#_partial_routes now explains it isn't but I think it is worth repeating it in the docstring.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -839,7 +839,7 @@
      The options are:
 
      `:router-targets` - (REQUIRED) A *vector* of ui components that are router targets. The first one is considered the \"default\"
-     (purely for the purpose if initial state; you always need to explicitly route to a particular target).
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -838,7 +838,8 @@
 
      The options are:
 
-     `:router-targets` - (REQUIRED) A *vector* of ui components that are router targets. The first one is considered the \"default\".
+     `:router-targets` - (REQUIRED) A *vector* of ui components that are router targets. The first one is considered the \"default\"
+     (purely for the purpose if initial state; you always need to explicitly route to a particular target).
```
awkay, line 842: Sorry, didn't see this PR till now for some reason. You have a typo: "if" should be "of"

holyjak: @awkay No problem! Fixed the typo now.

awkay:
👍

Issue #443, opened 2020-12-21
`comp/get-query` of component with `:use-hooks? true` returns `nil` in CLJ


fjolne:
```clojure
(defsc Comp [this props]
  {:query [:item]})

(defsc CompWithHooks [this props]
  {:query      [:item]
   :use-hooks? true})

;; CLJ (not CLJS)
(comp/get-query Comp) ;; => [:item]
(comp/get-query CompWithHooks) ;; => nil
```

awkay:
CLJ support should be considered very rough. I don't use it, and almost no one else does. I have no time for these tasks. If you want to address them, you're welcome to try...but for the moment I do not plan to fix.

fjolne:
sure, no problem, i’ll try to fix if it becomes crucial, just bookkeeping.

MrEbbinghaus:
I think this is because `comp/add-hook-options!` is not implemented for clj

awkay:
Well, if you really think about the whole thing use-hooks should proably just be ignored in CLJ...there's no hooks API, after all. The fix is to just make the macro ignore the option for CLJ I think.

PR #446, opened 2020-12-31
Issue #443: fixed comp/get-query for ':with-hooks? true' in CLJ

fjolne:
The problem was in CLJ component class not having `comp/component-options`. Treating hooks components as a regular React class in CLJ fixes the issue.


awkay:
Cool. Thanks!

Issue #444, opened 2020-12-21
`comp/fragment` node throws on rendering in CLJ


fjolne:
```clojure
(require '[com.fulcrologic.fulcro.dom-server :as dom])

(defsc Child [this props]
  (comp/fragment (dom/div "one")))
(def ui-child (comp/factory Child))

(defsc Root [this props]
  (dom/div (ui-child {})))

(dom/render-to-str ((comp/factory Root) {})) ;; throws java.lang.IllegalArgumentException: Key must be integer
```

Seems to only happen when the `comp/fragment` is the top form in the child and the parent wraps it in some HTML tag.

PR #447, opened 2020-12-31
Issue #444: fixed comp/fragment rendering in CLJ

fjolne:
I'm not very sure about this one, tests are passing, but would like to test on a prod codebase.


fjolne:
Tested on real codebase, ready to merge.

PR #453, opened 2021-01-18
 Hooks: shared props fix

fjolne:
Shared props are incorrectly bound to dynamic var in hooks render fn, which leads to issues with e.g. i18n.


awkay:
Shoot. Thanks for seeing that. I reverted a hot-code reload fix for initLocalState, and that seems to now be in your branch but not develop :(

Would you mind resetting your branch to develop and re-pushing the change?  If not, I can just hand code it in.

fjolne:
Sure, done.

awkay:
Thanks!

awkay:
Released in 3.4.13

PR #454, opened 2021-01-25
[WIP] propagate shared props to mutation action handlers

fjolne:
i18n doesn't work in mutations because bindings aren't propagated. Tests which check the defmutation macro output do NOT pass: it's somewhat tricky to make them work with gensym, and anyway would like to know what you think about the problem and solution in general.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/mutations.cljc
+++ b/src/main/com/fulcrologic/fulcro/mutations.cljc
@@ -436,10 +436,13 @@
                                     (let [action? (str/ends-with? (str handler-name) "action")]
                                       (into acc
                                         (if action?
-                                          [(keyword (name handler-name)) `(fn ~handler-name ~handler-args
-                                                                            (binding [comp/*after-render* true]
-                                                                              ~@handler-body)
-                                                                            nil)]
+                                          [(keyword (name handler-name)) `(fn ~handler-name [env#]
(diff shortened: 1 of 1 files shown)


awkay:
Yeah, so, I've been giving that some thought lately. The i18n stuff was designed with UI in mind, and there is no intention of them working in mutations, or in the back-end really (other than SSR).  The shared props are also not necessarily the ideal mechanism, since that ties their use to Fulcro, and most of the library is quite usable in just about any context.

The Fulcro integration of having the translations be loadable is nice, but it is also a trivial thing to implement. It's just a single request that puts them somewhere, and perhaps the state atom isn't the best overall place because it again ties the user to Fulcro.

There is also plenty of reason to want to use them on the back-end, and have string extraction there (see my Java library) https://github.com/awkay/easy-i18n, which could easily be used from Clojure, optionally with a wrapper.

I'm tempted to combine the solutions.

So, on the back-end, you want to set the locale for a request, and on the front-end you usually only care what the current locale is.

So, I'm thinking the better approach would be a dynamic var for locale, which in cljs could have a root binding that gets set, and in CLJ you could do `binding` on the request.

That's not a bw compatible solution per-se, but I think it's a better overall solution.

awkay:
The point of shared is really more about the UI tree. I understand it would be nice for i18n to have this, but I don't think I want to go down this rabbit hole here.

PR #459, opened 2021-02-16
Replace fn? app-specs with ifn?

MrEbbinghaus:
E.g. a keyword should be a valid fn for `:shared-fn` (and other config keys)

Like in the example for fulcro-i18n
`:shared-fn ::i18n/current-locale`

In the current state this floods the console with errors from guardrails.

image


awkay:
Hm...not all of those, no.  `:shared-fn` I guess *could* make sense, but `tx!` is a function that takes an app-ish and txn and side-effects...that is definitely not an `ifn?`.

awkay:
Thanks :)

PR #460, opened 2021-02-17
annotating dom where possible with data-fulcro-source html attribute to show namespace and line number

Pancia:
Got it mostly working, but expressions/symbols/js-objects are too tricky for not a lot of gain...
Also I think dom macro spec puts stuff under attrs as an expression, instead of realizing there is no attrs map.
See `src/workspaces/com/fulcrologic/fulcro/cards/source_annotation_cards.cljs`

### DIFF
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -2,12 +2,13 @@
  :nrepl    {:port 9000}
  :dev-http {9001 "resources/public"}
  :jvm-opts ["-Xmx2G"]
- :builds   {:workspaces {:target     nubank.workspaces.shadow-cljs.target
-                         :ns-regexp  "-cards$"
-                         :output-dir "resources/public/js/workspaces"
-                         :asset-path "/js/workspaces"
-                         :preloads   [com.fulcrologic.fulcro.inspect.websocket-preload
-                                      com.fulcrologic.fulcro.inspect.dom-picker-preload]}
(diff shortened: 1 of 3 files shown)


Pancia:
I figured out in f31ec2d the ability to annotate dom elements when the attribute map is a symbol or expression.


I also understood better what is happening that is causing some elements to not get any source annotation.
If you have an expression like `(div (p "broken"))`, the p element is parsed by the spec 
```
(s/def ::dom-macro-args
  (s/cat
    :css (s/? keyword?)
    :attrs (s/? (s/or :nil nil?
                  :map ::map-of-literals
                  :runtime-map ::map-with-expr
                  :js-object #(instance? JSValue %)
                  :expression list?
                  :symbol symbol?))
    :children (s/* (s/or :string string?
                     :number number?
                     :symbol symbol?
                     :nil nil?
                     :list sequential?))))
```
to be an `:expression` attribute, even though it should more correctly be under `:children`.
However the code seems to account for this later in the function `emit-map`, and it clearly must work since fulcro is in use.


The problem is that we essentially end up needing runtime information (does a form evaluate to a map or an element ?) at compile time.

However:
1. Annotating nodes with source isn't crucial to have on every element.
2. You can help the macro by simply adding to the previous example an empty attribute map `(div {} (p "fixed"))`.

PR #465, opened 2021-03-06
Check query is valid at dev runtime

holyjak:
to provide a better error message than "Invalid expression " from EQL.

TODO:
* also for set-query!
* get the check working upon refresh

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
@@ -156,11 +156,11 @@
   (though the Spec it uses might be both more strinct / more lax than the code
   so do not take it absolutely.
   "
-  [query]
+  [query component-name-fn]
   (when (and #?(:clj true :cljs goog.DEBUG)
              query
              (false? (s/valid? ::eql/query query)))
     (log/error (str "The composed root query is not valid EQL. The app may crash. See `(comp/get-query "
-                    (some-> query meta :component comp/component-name) ")`")
+                    (some-> query meta :component component-name-fn) ")`")
                query))
   query)
--- a/src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
@@ -150,3 +150,17 @@
     (if (vector? <>)
       <>
       [<>])))
+
+(defn- check-query-valid
+  "Runtime check that the (typically root) query looks valid
+  (though the Spec it uses might be both more strinct / more lax than the code
+  so do not take it absolutely.
+  "
+  [query]
+  (when (and #?(:clj true :cljs goog.DEBUG)
+             query
+             (false? (s/valid? ::eql/query query)))
+    (log/error (str "The composed root query is not valid EQL. The app may crash. See `(comp/get-query "
+                    (some-> query meta :component comp/component-name) ")`")
+               query))
+  query)
--- a/src/main/com/fulcrologic/fulcro/application.cljc
+++ b/src/main/com/fulcrologic/fulcro/application.cljc
@@ -186,9 +186,9 @@
   (when (and #?(:clj true :cljs goog.DEBUG)
              query
              (false? (s/valid? ::eql/query query)))
-    (throw (ex-info (str "The composed root query is not valid EQL. See `(comp/get-query "
-                         (some-> query meta :component comp/component-name) ")`")
-                    {:query query, :root (some-> query meta :component)})))
+    (log/error (str "The composed root query is not valid EQL. The app may crash. See `(comp/get-query "
+                    (some-> query meta :component comp/component-name) ")`")
+               query))
   query)
 
 (defn initialize-state!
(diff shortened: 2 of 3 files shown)


src/main/com/fulcrologic/fulcro/application.cljc
```diff
@@ -181,13 +182,23 @@
                                       (= "ui" ident-ns)
                                       (str/starts-with? ident-ns "com.fulcrologic.fulcro.")))))))))
 
+(defn- check-root-query-valid [query]
+  (when (and #?(:clj true :cljs goog.DEBUG)
+             query
+             (false? (s/valid? ::eql/query query)))
+    (throw (ex-info (str "The composed root query is not valid EQL. See `(comp/get-query "
```
awkay, line 189: Do not throw. I'd much rather see an error in the console log that has no stack trace, since the S.T. is useless. Also, if the check is wrong and the query right for some reason (bug) it won't break things.

awkay:
You asked about RAD demo refresh. There are two things going on there

https://github.com/fulcrologic/fulcro-rad-demo/blob/develop/src/shared/com/example/client.cljs#L44

A refresh of dynamic queries and a re-mount. The query for the UI refresh is in the renderer (refresh calls the installed renderer...see keyframe-render). Each renderer uses `db->tree`.  

For this check, though, I think just `mount!` is the right place, probably right about here:

https://github.com/fulcrologic/fulcro/blob/develop/src/main/com/fulcrologic/fulcro/application.cljc#L396

holyjak:
Hm, this does not work either. The reason surely is this warning logged:

> react_devtools_backend.js:2430 error when calling lifecycle function com.example.client/refresh 

i.e. the app crashes before it gets to the point in moun! So I need to find out where does this happen.

Question 2: To be able to check also `comp/set-root!` I need the `check-root-query-valid` fn to be visible both to `.application` and `.components`. So I could move it into the comp ns but I also want to keep it private not to confuse the users and increase the API surface - but then app. could not see it. Possible solutions:

1. Bypass private and call it anyway (I believe this is possible)
2. Duplicate it in both ns
3. Introduce a new ns, `com.fulcrologic.fulcro.impl` and put it there

Thoughts?

awkay:
There's already a namespace called do-not-use that is usable from both components and application. Use that.  It's mostly a dumping ground for things that were legacy and had no place, and might possibly get removed. So, adding it there won't hurt anything, and anyone that uses it is doing so at their own risk.

awkay:
It is quite common to put something like `[df/marker-table '_]` in a component's query, but since the `df/marker-table` is a symbol (as is `'_`, it causes the check to be skipped.

So, I think we should assume that any symbols in vectors that have count 2 are fine, and we should elide such idents from the query and check the rest. NOTE: This also could be on a join, so it is probably best to just replace them with a placeholder keyword.

Thus:

`{[table-sym some-id] (comp/get-query Component)}` would become `{:ident-placeholder [:query-for-Component]}`, etc.

src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
```diff
@@ -150,3 +150,17 @@
...
       <>
       [<>])))
+
+(defn- check-query-valid
+  "Runtime check that the (typically root) query looks valid
+  (though the Spec it uses might be both more strinct / more lax than the code
+  so do not take it absolutely.
+  "
+  [query component-name-fn]
+  (when (and #?(:clj true :cljs goog.DEBUG)
```
holyjak, line 160: should there be `:clj true` or false?

holyjak:
I guess the last comment actually belongs to #446? 

I think this is ready to merge, as long as you thing that defaulting to doing the check in clj is ok (`:clj true`).

src/main/com/fulcrologic/fulcro/algorithms/do_not_use.cljc
awkay: doesn't really matter. Probably false, since you wouldn't run an SSR on a broken app, but whatever :D

src/main/com/fulcrologic/fulcro/application.cljc
```diff
@@ -17,6 +17,7 @@
     [com.fulcrologic.fulcro.inspect.inspect-client :as inspect]
     [edn-query-language.core :as eql]
     [clojure.string :as str]
+    [clojure.spec.alpha :as s]
```
awkay, line 20: It says you added this, but I don't see where you use it.

src/main/com/fulcrologic/fulcro/components.cljc
```diff
@@ -1091,6 +1091,7 @@
                      (string? class-or-factory) class-or-factory
                      (some-> class-or-factory meta (contains? :queryid)) (some-> class-or-factory meta :queryid)
                      :otherwise (query-id class-or-factory nil))]
+    (util/check-query-valid (get-query class-or-factory) component-name)
```
awkay, line 1094: you have to pass the state map to get query to actually see the new dynamic query

`(get-query c-of-r @state-atom)`

src/main/com/fulcrologic/fulcro/application.cljc
```diff
@@ -393,7 +395,8 @@
                                        ::root-factory root-factory
                                        ::root-class root)
                                      (update-shared! app)
-                                     (indexing/index-root! app)
+                                     (util/check-query-valid (comp/get-query root) comp/component-name)
```
awkay, line 398: Same as the other comment. `get-query` on initial mount is probably just the static query declared on components, but on remount it needs to be the current dynamically bound query, which requires you pass the state map to `get-query`.

awkay:
I'm fixing the rest of this myself, because now is when I have time to deal with it.

awkay:
Merged manually.

PR #469, opened 2021-04-04
Outline some changes for fixing JVM clojure compilation

vemv:
Hi!

Because of an issue reported in Eastwood, I tried to run that linter over this project.

That made me find out that despite that the Fulcro project is coded in .cljc, a few spots can't be successfully compiled over the JVM.

This PR outlines a few possible fixes. I only checked over `main`, not `test` or others.

It's not intended for merging as I'm not a primary user of Fulcro. If you are interested in this improvement you could simply pick it up and improve it as desired.

In case it helps, here is the alias I used for checking valid compilation:

```clj
           :lint/eastwood
           {:extra-deps {jonase/eastwood {:mvn/version "0.4.0"}}
            :jvm-opts   ["-Dclojure.main.report=stderr"]
            ;; feel free to add "src/test" and others:
            :main-opts  ["-m" "eastwood.lint" "{:source-paths,[\"src/main\"]}"]}
```

...a valid compilation will simply return linter faults, and an invalid compilation will throw an exception due to the cljs-only bits.

(another way of checking would be using tools.namespace `refresh`. Or use something akin to `lein check` if that exists for deps.edn?)

Hope it helps!

-V

### DIFF
--- a/src/main/com/fulcrologic/fulcro/alpha/raw_components.cljc
+++ b/src/main/com/fulcrologic/fulcro/alpha/raw_components.cljc
@@ -75,6 +75,7 @@
     (inspect/app-started! app)
     app))
 
+#?(:cljs
 (defn factory
   "A Fulcro component factory for RAW React usage.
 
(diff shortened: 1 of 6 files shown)


awkay:
Thanks for trying to help, some of these nses are alpha and will be going away again. You've tried adding :cljs to cljs files in some case, and in other cases the suggestions would actually break code. I'm sorry, I'm not interested in these kinds of patches (linter says it's bad, therefore you send me patches). 

If you have a reproducible problem that affects real software please report it as an issue.

PR #470, opened 2021-04-05
Added index.html

alexparens:
https://github.com/fulcrologic/fulcro/commit/9e0612d5d031155154318bc2dac47ae605254773

In the commit above, workspaces.html was renamed to index.html. However, it seems that workspaces.html was only deleted and there was no index.html to access the workspaces.

This pull request simply adds the index.html with the content from the previous workspaces.html. Have tested to confirm that workspaces can be accessed properly through localhost:9001. Thank you for reviewing.

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Fulcro Workspaces</title>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


awkay:
oopos. Thanks. Must have been on my ignore list.

awkay:
oh, actually, I just plain forgot to commit it!

PR #476, opened 2021-04-23
turn `m/returning`s optional arg into a generic opts map

tylernisonoff:
more work will be required to support `without` due to circular dependencies with public functions.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/mutations.cljc
+++ b/src/main/com/fulcrologic/fulcro/mutations.cljc
@@ -331,11 +331,14 @@
   `env` - The env of the mutation
   `class` - A component class that represents the return type.  You may supply a fully-qualified symbol instead of the
   actual class, and this method will look up the class for you (useful to avoid circular references).
+  `opts` (optional):
+   - `query-params` - Optional parameters to add to the generated query
 
   Returns an update `env`, and is a valid return value from mutation remote sections."
   ([env class]
    (returning env class nil))
-  ([env class query-params]
+  ([env class {:keys [query-params]
+               :as   opts}]
    (let [class (if (or (keyword? class) (symbol? class))
                  (rc/registry-key->class class)
                  class)]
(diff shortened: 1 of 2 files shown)


src/main/com/fulcrologic/fulcro/mutations.cljc
```diff
@@ -331,11 +331,14 @@
...
   `class` - A component class that represents the return type.  You may supply a fully-qualified symbol instead of the
   actual class, and this method will look up the class for you (useful to avoid circular references).
+  `opts` (optional):
+   - `query-params` - Optional parameters to add to the generated query
 
   Returns an update `env`, and is a valid return value from mutation remote sections."
   ([env class]
    (returning env class nil))
-  ([env class query-params]
+  ([env class {:keys [query-params]
```
wilkerlucio, line 340: the problem here is that this a breaking change, users of query params will have unexpected results on upgrade. If we need more things here, better to add another argument or make a new function.

tylernisonoff: @wilkerlucio this branch hasn't been released, so its not breaking yet!
Tony and I were discussing here: https://clojurians.slack.com/archives/C68M60S4F/p1619214744201500

tylernisonoff: (For clarity, the `query-params` parameter was added after the last public release of fulcro)

wilkerlucio: ah, ok, my bad then, sorry for taking conclusions looking only in the small

tylernisonoff: No problem, I should/could have linked to the conversation in the PR.

awkay:
We can move any implementation towards a dependency leaf and just create a ref from the old location to the new. For now, though, this at least fixes the public API.

PR #478, opened 2021-04-24
Raw API: Fix batched notifications, hooks returning nil & more.

LuisThiamNye:
- Changes to `use-root` and `use-component` hooks:
  - The component state is initialised on the first mount of the component, then current-props are initialised and ready to use on the first render
  - An effect hook manages the addition/removal of the render listener
  - use-component:
    - Randomly generates a UUID to use for the render listener so that multiple components can share the same ident
    - Uses the :ident option if provided
- Changes to raw.application
  - In `render!` — use :batch-notifications (not :batch-renders) to match the key provided to the application.
  - In `add-component!` and `add-root!` — Refactor out parts into helper functions which are also used by the hooks. (Two of these are in raw.components)
  - `add-component!` uses the :ident option if provided, and `initialize?` is now true by default
- Replaced references to components/*after-render* with raw.components/*after-render*

### DIFF
--- a/src/main/com/fulcrologic/fulcro/raw/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/raw/components.cljc
@@ -797,6 +797,21 @@
            ast-nodes (-> query eql/query->ast :children)]
        (get-subquery-component* component ast-nodes query-path)))))
 
+(defn get-traced-props
+  "Uses `fdn/traced-db->tree` to get the props of the component at `ident`. If `prior-props` are not stale,
+   those are returned instead."
+  [state-map component {:keys [ident prior-props]}]
+  (let [query (get-query component state-map)]
+    (if (fdn/possibly-stale? state-map prior-props)
+      (fdn/traced-db->tree state-map ident query)
+      prior-props)))
+
+(defn has-active-state?
+  "Returns true if there is already data at a component's ident"
+  [state-map ident]
+  (let [current-value (get-in state-map ident)]
+    (and (map? current-value) (seq current-value))))
+
 (comment
   (def Person (entity->component
                 {:person/id        1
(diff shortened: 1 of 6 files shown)


src/main/com/fulcrologic/fulcro/raw/components.cljc
```diff
@@ -797,6 +797,21 @@
...
+(defn get-traced-props
+  "Uses `fdn/traced-db->tree` to get the props of the component at `ident`. If `prior-props` are not stale,
+   those are returned instead."
+  [state-map component {:keys [ident prior-props]}]
+  (let [query (get-query component state-map)]
+    (if (fdn/possibly-stale? state-map prior-props)
+      (fdn/traced-db->tree state-map ident query)
+      prior-props)))
+
+(defn component-exists?
```
awkay, line 809: Not an ideal name, but I'm not sure what to call it. There's already a bit of naming screw-up around the word "component" peppered throughout the codebase, so I'd prefer not to make that worse :D   I'll give it a bit of thought. Suggestions?

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -686,7 +687,7 @@
                         (uism/trigger! app router-id :route! event-data))
                       ;; make sure any transactions submitted from the completing action wait for a render of the state machine's
                       ;; startup or route effects before running.
-                      (binding [comp/*after-render* true]
+                           (binding [rc/*after-render* true]
```
awkay, line 690: Could you add a `;; TASK: check on after-render` above this line?

src/main/com/fulcrologic/fulcro/raw/components.cljc
LuisThiamNye: I agree. Some ideas from the top of the head:

- loaded?
- loaded-state?
- has-some-state?
- some-state?
- state-exists?
- occupied?
- ident-occupied?
- in-use?

`has-active-state?` is already good and descriptive though.

I now wonder whether this is the right function, or if it is even necessary.

Is the `map?` check necessary? Or would it be sufficient to simply use `(seq (get-in state-map ident))`?

awkay: Yeah, has-active-state? is fine by me.  And yes, should probably check it is a non-empty map so things don't get confused. That's the only way it will actually work.

LuisThiamNye: Very well, I have changed the name to `has-active-state?`

awkay:
Cool. If it'll probably evolve a bit more, but this is a good start, thanks.

PR #480, opened 2021-05-12
Add meta to computed-factory

JJ-Atkinson:
* Fix `computed-factory` such that `(comp/get-query computed-factory ..?)` works


awkay:
good catch

PR #481, opened 2021-08-20
Parameterise response in handle-api-request

alilee:
Allows you to start building the response from a known non-empty state (eg. with the requests :session), which allows augment-handlers to relax their assumptions about any other handler's changes, and the potential for clashes.

Hi Tony, this is just a sketch - I haven't run the regression tests. Cheers, Alister.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/server/api_middleware.clj
+++ b/src/main/com/fulcrologic/fulcro/server/api_middleware.clj
@@ -63,8 +63,12 @@
   "Given a parser and a query: Runs the parser on the query,
    and generates a standard Fulcro-compatible response, and augment the raw Ring response with
    any augment handlers that were indicated on top-level mutations/queries via
-   `augment-response`."
+   `augment-response`.
+   Optionally, specify the starting response map to be fed to the augment handlers (ie. to
+   include the request `:session`."
   [query query-processor]
+  (handle-api-request query query-processor {})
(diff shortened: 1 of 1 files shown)


awkay:
I thought you wanted to start with a different base on line 60. I think that would be a more appropriate thing to support. What you're doing here isn't really very interesting I think.

alilee:
Exactly - I was lost...

I think the obvious downside of this is that the session now gets updated every request, rather than just when :session is present, which might be unwanted on a persistent/database session. Wrap-session would then need to diff before it makes the persistent update.

src/main/com/fulcrologic/fulcro/server/api_middleware.clj
```diff
@@ -56,8 +56,10 @@
   `augment-response`. Runs each in turn and accumulates their effects. The result is
   meant to be a Ring response (and is used as such by `handle-api-request`."
   [response]
```
awkay, line 58: Missing parens.

awkay:
Right, I'm not sure any of this makes sense.  If you want a specific way of doing it, you can always write your own lines of code to handle the request...it's a tiny amount of code.

alilee:
Yeah, it's not a worthy case. It can only be any kind of problem where: there are augment-handlers working on more than one part of the session AND they can't know about each other, in which case you tweak those routines for your app.

PR #483, opened 2021-09-01
Put kondo exports on the classpath

holyjak:
so that apps depending on Fulcro can actually import them

See https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting-and-importing-configuration


holyjak:
I have verified that I can `clj-kondo --copy-configs` and that kondo thereafter stops reporting some problems.

deps.edn
```diff
@@ -1,4 +1,7 @@
-{:paths   ["src/main"]
+{:paths   ["src/main"
+           ;; include clj-kondo exports in the JAR so that they can be imported,
+           ;; see https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting-and-importing-configuration
+           "clj-kondo-exports"]
```
awkay, line 4: I'd prefer this be `src/clj-kondo`

holyjak: renamed

PR #486, opened 2021-09-04
Feat/484 autoadd tempid2mut joins

### LINKED ISSUE #484: Fix templates and book to recommend Pathom setting for always returning :tempids
Add :tempids to mutation join queries by default so that all users do not need to do that themselves (and '* if there was no query before to preserve Pathom's behavior of everything the whole mutation output if the user did not ask for anything in particular).

Add ::p/errors to all queries and mutations - most users of Fulcro use Pathom and need this and an extra keyword will not harm those who don't (and they can always remove it via a custom global transform).

Also expand the default elision [list to RAD's](https://github.com/fulcrologic/fulcro-rad/blob/7db67fbab7211dcaa87658fb35c0165d7eb322fa/src/main/com/fulcrologic/rad/application.cljc#L25) (for the sake of simplicity, include also the \*.rad.\* ones, they are just keywords so do not create any dependency on RAD and will make its life easier).

## Consequences

* Many users likely do some of this in their global eql transforms. If they combine their with Fulcro, they might end up with these props twice in the query, but that should not harm anything, since we get back just a map, so no duplication there
* Users will far less run into problems where tempid remapping does not work, b/c they used a mutation join and forgot to add :tempids
* Users will get pathom errors back by default, which will make it easier to discover what went wrong with their query/mutation

holyjak:
Fixes #484

### DIFF
--- a/src/test/com/fulcrologic/fulcro/application_spec.cljc
+++ b/src/test/com/fulcrologic/fulcro/application_spec.cljc
@@ -2,12 +2,15 @@
   (:require
     [fulcro-spec.core :refer [specification provided! when-mocking! assertions behavior when-mocking component =>]]
     [clojure.spec.alpha :as s]
+   [clojure.walk :as walk]
     [edn-query-language.core :as eql]
     [com.fulcrologic.fulcro.specs]
     [com.fulcrologic.fulcro.components :as comp]
     [com.fulcrologic.fulcro.routing.dynamic-routing :as dr]
     [com.fulcrologic.fulcro.ui-state-machines :as uism]
     [com.fulcrologic.fulcro.application :as app :refer [fulcro-app]]
+   [com.fulcrologic.fulcro.raw.application :as rapp]
+   [com.wsscode.pathom.core :as p]
     [clojure.test :refer [is are deftest]]))
 
 (deftest application-constructor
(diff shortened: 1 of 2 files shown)


holyjak:
Draft because I have to fix the tests but the code is I believe finished.

holyjak:
OMG, Calva has messed up with formatting in a number of files. I have to undo that :(

holyjak:
When I print the argument to `global-eql-transform`, I get this example:

```clojure
{:type :root, 
 :children [{:type :join, :dispatch-key :i-fail, :key :i-fail, :query [*], :component anonymous,
                  :children [{:dispatch-key *, :key *}]}]}
```

but then the code that should add `:tempids` to mutations in https://github.com/fulcrologic/fulcro/issues/484#issuecomment-911967463 is confusing because it seems to assume it does not see the `ast` itself but an individual child of it. I assume that I have just misunderstood how it is used???

awkay:
So my implementation in RAD for GEQLT is sort of a "bug waiting to happen" that relies (I *think) on an internal detail of how the transform is used in tx-processing. For mutations, which each can have their own remote, the tx processing breaks apart the AST for each mutation to be sure it can dispatch them individually to different remotes (if necessary). Thus, then it calls GEQLT on a *mutation* is does send just the single node (please verify that with debug logging in a real app...this is from dusty memory).

A *query*, however, does not get broken apart because `load!` won't let you use more than one explicit remote per load. So, you get the whole tree of AST nodes.

When the split mutation nodes reach the *network* layer, then IF they all came from the same top-level transact AND they all go to the same remote *then* those nodes are recombined for a single network send.

So, technically you should make GEQLT *expect* that it might get any kind of valid AST (root node or single child node). The elision code does this, and will work for either case. The mutation conditions are making a convenient assumption, but you're right that to prevent future bugs it should also be tolerant of being handed a root AST node. That way an alternate tx-processing plugin can use it without this internal dependency.

holyjak:
You are right, for an actual mutation I get printed

```clojure
{:dispatch-key com.example.mutations/create-random-thing, :key com.example.mutations/create-random-thing, :params {:tmpid #fulcro/tempid["754b7b97-29d8-435f-b4ae-22260ac85445"]}, :type :call}
```

But if we want to free the GEQLT from depending on this implementation detail, how best to do that? Some options:
1. Use clojure/walk as I do now - simple somewhat wasteful (we know that we only need to descend from :root to its children or if the thing we are processing is a join - and this max once, i.e. not into a join inside a join - but walk walks all the nodes down to the leaves). Another downside can be that we bring in hitherto unused ns, making the .js bigger
2. Use zippers to only descend where we need it - likely less wasteful but more complicated and can also make .js bigger
3. DIY - manually check for :root / mutation join and proceed accordingly

awkay:
I'd probably do a simple check and recurse one level. Mutations will never be deeper than that.

src/test/com/fulcrologic/fulcro/application_spec.cljc
```diff
@@ -25,22 +32,103 @@
...
+         :message "Found values missing from the set"}))))
+
+;; (letfn [(transform [query] (sort-nested (transform-query query)))]
+;;   (transform '[(mutation6)]))
+
+(specification "Default EQL transform additions"
+  (behavior "Adds ::p/errors to all queries BUT not mutations"
+    (letfn [(transform [query] (sort-nested (transform-query query)))]
+      (assertions
+        (transform [:a]) => [:a ::p/errors]
```
awkay, line 102: These tests would be way clearer if you had a helper that took EQL, turned it into an AST, passed it to transform, then converted it *BACK* to EQL...then you could write something like

```
"Leaves regular mutations alone"
(transform '[(mutation)]) => '[(mutation)]
"Adds :tempids to mutation joins"
(transform '[{(mutation)] [:x]}) => '[{(mutation) [:x :tempids]}]
```

holyjak:
Sorry, I am not following. Isn't it exactly what transform-query is doing?

The problem is the result contains more than I want to test in the particular test...

awkay:
Decided to take a different route.

PR #488, opened 2021-10-03
docs fix: deferred-timeout default seems to be 20ms

holyjak:
According to https://github.com/fulcrologic/fulcro/blob/f1ff70f862c46b8dd10e886686271916cdfa590a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc#L329 and https://github.com/fulcrologic/fulcro/blob/f1ff70f862c46b8dd10e886686271916cdfa590a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc#L676 the defaults seems to be actually 20ms, not 100ms as the docs claim.


awkay:
Hmn...actually, I thin kthis needs a doc fix instead :D 

I think 100ms felt too slow and I tuned it back to 20 but forgot I documented it...perhaps just remove it from the docs?

PR #492, opened 2021-11-08
Fix and improve `>defn` clj-kondo hook

MrEbbinghaus:
The current clj-kondo hook for `>defn` produces a false positive when using argument predicates ("such that").

This would fail because the current implementation counts everything before `=>`:
```clojure
(>defn my-fn [x]
  [::x | #(pos? x) => any?] 
  x)
```

-------------

I took the chance and added some more checks to the gspec validation.

gspec reference:
`[arg-specs* (| arg-preds+)? => ret-spec (| fn-preds+)? (<- generator-fn)?]`

What errors are caught?
- Missing `=>` 
- Missing `ret-spec`
- `|` present but no `arg-preds`
- number of `arg-specs` is not equal to number of arguments

@holyjak Maybe you want to review this as well?


MrEbbinghaus:
Maybe we should move this to [fulcrologic/guardrails](https://github.com/fulcrologic/guardrails) .. ?

awkay:
> Maybe we should move this to [fulcrologic/guardrails](https://github.com/fulcrologic/guardrails) .. ?

If you're up for it, yes please :)

PR #493, opened 2021-11-10
Feat/log once

holyjak:
Add log functions that throttle the log (by arguments) so that the developer is not swamped by many occurrences of the same error.


src/main/com/fulcrologic/fulcro/algorithms/do_not_use/log_once.cljc
```diff
@@ -0,0 +1,18 @@
...
+  (:require
+    [taoensso.encore :as encore]
+    [taoensso.timbre :as log]))
+
+; TODO Should we limit the size and use a LRU cache or st.? Hopefully not really an issue in a webapp?
+; Or implement a DIY TTL so we repeat the error sometimes? (=> {<args> (js/Date.now)})
+
+;(encore/memoize 60000 (constantly nil))
+
+(defn error-once! [& args]
```
awkay, line 17: how about `log-once!` with level as first arg. The encore/memoize is sufficient.

src/main/com/fulcrologic/fulcro/algorithms/normalize.cljc
```diff
@@ -60,16 +62,13 @@
                           (when-not (empty? is)
                             (swap! refs
                               (fn [refs]
-                                (reduce (fn [m [i x]]
-                                          (update-in m i merge x))
-                                  refs (zipmap is xs)))))
+                                (reduce (fn [m [i x]] (update-in m i merge x)) ; merge2data-entity
+                                  refs (map vector is xs)))))
```
awkay, line 66: I don't want changes to this file. You can add a docstring if you want, and comments would be OK *if* they are clear to the reader....having zipmap change to map is unacceptable. I don't want any algorithmic changes at all.

I'd be ok with var renames, but not ones that alias outer things or collide with clojure.core names. So, for example, I'd be ok with changing `k` to `join-key` and `sel` to `subquery`.

awkay: See my updated comment to the issue. I think that's actually the way to go (a function that calculates error after thrown)

src/main/com/fulcrologic/fulcro/algorithms/do_not_use/log_once.cljc
awkay: Thought I already made this comment somewhere, but it would be sufficient to have an arity 2 macro for log-once that elides with the same logic as timbre. Perhaps 2 macros: warn-once! and error-once!. The fixed arity then makes it possible for the macro to use a memoized (private) function.

awkay: Also, let's convert timbre-support to a CLJC namespace, and put this stuff in there. These should be usable by libraries, and `do-not-use` really is meant to be something no one uses...though I know that isn't necessarily true :D

src/main/com/fulcrologic/fulcro/algorithms/timbre_support.cljc
```diff
@@ -0,0 +1,100 @@
...
+  See the development_preload.cljs and shadow-cljs.edn files in the latest Fulcro 3 template for an example.
+  "
+(:require
+  [clojure.string :as str]
+  [taoensso.encore :as encore]
+  [taoensso.timbre :as log]))
+
+;; Taken mostly from timbre itself.  Modified to output better results from ex-info exceptions (e.g. to improve expound
+;; experience)
+#?(:cljs
```
awkay, line 41: In general put these in the body in case it is used in a CLJC file, so that the function is at least defined. Then note in docstring that it is only implemented for cljs.

src/main/com/fulcrologic/fulcro/algorithms/timbre_support.cljc
```diff
@@ -0,0 +1,100 @@
...
+   console appender from this namespace to get better logging output in cljs."
+  ([data] (prefix-output-fn nil data))
+  ([opts data]                                              ; For partials
+   (let [{:keys [level ?ns-str ?file ?line]} data]
+     (str (str/upper-case (name level)) " " "[" (or ?ns-str ?file "?") ":" (or ?line "?") "] - "))))
+
+(defn- warn-once!* [message] (log/warn message))
+(defn- error-once!* [message] (log/error message))
+
+(def warn-once! "Log a warning but max once / 60s for a particular message" (encore/memoize 60000 warn-once!*))
```
awkay, line 99: I'd increase the timeout to at least 5 mins (300000)

src/main/com/fulcrologic/fulcro/algorithms/do_not_use/log_once.cljc
```diff
@@ -0,0 +1,18 @@
...
+  (log/log! :error :p args #_{:?line (fline &form)}))
```
awkay, line 18: Not sure why this change is still here???

awkay:
If you use timbre's `-elide?` method in a macro, as shown here https://github.com/ptaoussanis/timbre/blob/7bb3d648e1a49cf835233785fe2c07e43b2395da/src/taoensso/timbre.cljc#L692 

then then can be removed during compile to eliminate their overhead.

holyjak:
So st. like this:

```clojure
(def ^:private debug-once!* (encore/memoize five-minutes-ms warn))
(defmacro debug-once!
  "Log a debug msg but max once / 5 min for particular arguments. Contrary to timbre/debug, takes max 3 arguments"
  [& args]
  (when-not #?(:clj (timbre/-elide? :debug *ns*) :cljs false)
    `(debug-once!* ~@args)))
```

?

awkay:
> So st. like this:
> 
> ```clojure
> (def ^:private debug-once!* (encore/memoize five-minutes-ms warn))
> (defmacro debug-once!
>   "Log a debug msg but max once / 5 min for particular arguments. Contrary to timbre/debug, takes max 3 arguments"
>   [& args]
>   (when-not #?(:clj (timbre/-elide? :debug *ns*) :cljs false)
>     `(debug-once!* ~@args)))
> ```
> 
> ?

So macros, unless you're doing bootstrapped cljs (which is way harder) are only used by CLJ, so it's typically better to wrap the macro in the conditional reader with just a `:clj` side.

```
#?(:clj
    (defmacro ...))
```

awkay:
Were you going to finish making macros for this?

holyjak:
Yes, jsut hadn't the time yet

holyjak:
NOTE to self: Timber v5.1.2 makes easier using it in libraries in the ways that we want - see https://github.com/ptaoussanis/timbre/blob/master/CHANGELOG.md#new-since-v512

Issue #491, opened 2021-11-05
Improve error messages when users supply bad data to normalization


holyjak:
If the developer makes a mistake and thinks she is getting a vector of maps and tries makes a join for it in the query (which happens easily as the data model and componens evolve during the initial explorations) then init will blow up in a way that is hard to troubleshoot, with the error

```
Error: conj on a map takes map entries or seqables of map entries
    ...
    at cljs$core$merge (core.cljs:9131)
```

When I run the same code in Clojure, the error is much better:

```
Execution error (ClassCastException) at com.fulcrologic.fulcro.algorithms.normalize/normalize*$fn$fn (normalize.cljc:64) . 
class java.lang.Character cannot be cast to class java.util.Map$Entry
```

This happens with code such as:

```clojure
(defsc Tag [_ _]
  {:query [:tag/id :tag/desc]
   :ident  :tag/id
   :initial-state {}}
  (dom/p ""))

(defsc Menu [_ _]
  {:query         [{:tags (comp/get-query Tag)}] ; <-- WRONG, tags are strings not maps
   :ident         (fn [] [:component/id :Menu])
   :initial-state {}}
  (dom/p ""))

(defsc Root [_ _]
  {:query [{:task-filters (comp/get-query Menu)}]
   :initial-state (fn [_] {:task-filters  {:tags ["Important" "Urgent"]}})}
  (dom/p "root"))
```  

So what I would love is a better error message and **I would be happy to provide the code** to make it possible. However I guess I would need some guidance as the code in question:

https://github.com/fulcrologic/fulcro/blob/c840483be0c635825b47b8f6434c98de093bf83c/src/main/com/fulcrologic/fulcro/algorithms/normalize.cljc#L64

is quite hard to understand without enough "domain knowledge". One option would be to wrap it in a try catch:

```clojure
  (try
    (update-in m i merge x)
    (catch #?(:clj Exception :cljs :default) e
      (when-not (map? x)
        (throw (ex-info (str "Your query includes a join on an attribute whose value is a vector of non-maps: " v) 
                 {:data v, :k k, :class class, :q q})))
      (throw e)))
```

but there is surely a better solution. (I also obviously do not know which info is most helpful to include in the error message. Or rather where to get the name of the component with the incorrect query and the attribute that should not have a join.)

awkay:
On second thought, how about this:

[...]

PR #494, opened 2021-11-11
Better error reporting in normalize*

holyjak:
First step on the way to fixing #491

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/normalize.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/normalize.cljc
@@ -104,6 +104,109 @@
                   (recur (next q) (assoc ret k v))))))
           ret)))))
 
+(defn- upsert-ident
+  "Insert or merge a data entity into a state table under the given `ident`.
+  Ex.: `(upsert-ident {} [:person/id 1] #:person{:id 1 :age 42}) => {:person/id {1 #:person{:id 1, :age 42}}}`"
+  [state ident entity-map]
+  (try
+    (update-in state ident merge entity-map)
+    (catch Exception e
+      (when-not (map? entity-map)
+        (throw (ex-info (str "Query join indicates the data should contain a data map but the actual data is "
+                             (pr-str entity-map)
+                             " Joined component's ident: " ident)
+                        {:ident ident, :data entity-map})))
+      (throw e))))
+
+(comment
+  (upsert-ident
+    {:x 1, :person/id {1 #:person{:id 1, :age 42}}}
+    [:person/id 1] "not a map!")
+  ())
+
+(defn- normalize-with-friendly-errors
+  "Like `normalize*` but with error logger. Possibly slower and buggy. Use to report errorrs after normalize* failed."
+  [query data tables union-seen transform]
+  ;; `tables` is an (atom {}) where we collect normalized tables for all components encountered during processing, i.e.
+  ;; we only return the "top-level keys" with their data/idents and all "tables" are inside this
+  (let [data (if (and transform (not (vector? data)))
+               (transform query data)
+               data)]
+    (cond
+      (= '[*] query) data
+
+      ;; union case
+      (map? query)
+      (let [class (-> query meta :component)
+            ident (get-ident class data)]
+        (if-not (nil? ident)
+          (normalize-with-friendly-errors (get query (first ident)) data tables union-seen transform)
+          (throw (ex-info "Union components must have an ident" {}))))
+
+      (vector? data) data                                   ;; already normalized
+
+      :else
+      (loop [q (seq query), ret data]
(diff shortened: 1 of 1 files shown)


holyjak:
I have improved the error messagefrom #491  to be the following:

> Query join indicates the data should contain a data map but the actual data is "Important" Joined component's ident: [:tag/id nil]

holyjak:
If anybody knows of any other cases where normalize fails with an unhelpful error message, pls let me know!

src/main/com/fulcrologic/fulcro/algorithms/normalize.cljc
```diff
@@ -18,74 +20,193 @@
...
+                  (recur (next q) ret)
+                  (recur (next q) (assoc ret k v))))))
+          ret)))))
+
+(defn- better-normalize* [query data tables union-seen transform]
+  (try
+    (normalize* query data tables union-seen transform)
+    (catch #?(:clj Exception :cljs :default) e
+      (normalize-with-friendly-errors query data tables union-seen transform)
+      (log/warn "Unexpectedly, normalize* failed but normalize-with-friendly-errors did not. Please report it (with the inputs) to Fulcro")
```
awkay, line 207: I would think the ideal thing to do would be report error, then issue a final error that says "normalize failed, see above errors" and maybe includes the query and data that was provided...but don't throw again. A stack trace is almost certainly non-helpful since it will come from the low-level transaction system, not your code.

holyjak: Ok! Pushed a new version.

I still throw an exception in `upsert-ident` just to fail fast (and get the info faster to the developer) but changed this place to only log.

awkay:
Where were we on this??? I thought it was merged already :)

holyjak:
No, I had not time to work on the suggested changes due to re:Clojure. I will get back to it ASAP.

holyjak:
Reproducing from Slack:

> [Tony, Nov 15th]: My apologies on the normalize one. I’m torn. If you were only after improving one particular case of potential error, then perhaps it does make more sense just to augment the original function. I’m concerned about breakage, so the zipmap should remain as-is, but I think it is possible to refactor in some error checking without breaking things.  I had assumed you wanted to do a bunch of things to check for various errors (not sure what those would be off the top of my head, it was just my assumption).  What’s your take on the fragility of just adding this one error message to the original function?

holyjak:
### Testing

Test code, to be inserted into `com.fulcrologic.fulcro.algorithms.normalize`:

```clojure
(comment
  (do
    ;; FIXME: Why does this often take > 5s to run? Seems caused by `log/error in `upsert-ident, even if it only logs "test" ?!
    (comp/defsc Tag [_ _]
      {:query [:tag/id :tag/desc]
       :ident :tag/id
       :initial-state {}}
      nil)

    (comp/defsc Menu [_ _]
      {:query [{:tags (comp/get-query Tag)}] ; <-- WRONG, tags are strings not maps
       :ident (fn [] [:component/id :Menu])
       :initial-state {}}
      nil)

    (comp/defsc Root [_ _]
      {:query [{:task-filters (comp/get-query Menu)}]
       :initial-state (fn [_] {:task-filters {:tags ["Important" "Urgent"]}})}
      nil)

    (tree->db Root {:task-filters {:tags ["Important" "Urgent"]}}))
  )
```

The output will be:

> ERROR - Normalize failed and no data will be inserted into the client DB. Error: Query join indicates the data should contain a data map but the actual data is "Urgent" Joined component's ident: [:tag/id nil] Target component: com.fulcrologic.fulcro.algorithms.normalize/Root Data:  {:task-filters {:tags ["Important" "Urgent"]}}
>
> `=> {}`

holyjak:
### What has been done

1. Throw a user-friendly exception instead of the one thrown previously
2. Moved `update-in ...` into the friendly named `upsert-ident`
3. Renamed locals to more meaningful names
4. Added info about the origin of the code so that we know it if/when we consider refactoring in the future

holyjak:
From the [history of my discussions with Tony](https://clojurians.slack.com/archives/D010CPAPAH1/p1636145990010300):

> tony.kay  21 h 59 Nov 5th
>
> So, it would be nice if normalize could continue. For example, if you got back `m` without the `update-in` applied you'd just skip bad data in that case so, I wonder if making some kind of guarded wrappers for assoc/assoc-in/update/update-in would be better
>
> Jakub Holý: For the developer it is "better" if it blows up, with a good error, b/c then s/he will notice the problem 
>
> Tony: I don't agree with that [an error blowing up the ui]. An error in the logs that explains why my data isn't in state is better to me. I don't need a stack trace into an algorithm I didn't even directly call. I dislike the "burn the UI down" behavior of react in general...so I'm not a fan of making it happen ever

PR #496, opened 2021-11-15
Fix #495: elm. picker handles err. boundaries

### LINKED ISSUE #495: eb/error-bondary breaks Fulcro Inspect's element picker
When I wrap the _body_ of a defsc in `com.fulcrologic.fulcro.react.error-boundaries/error-boundary` then the Element Picker from Fulcro Inspect - Element will start ignoring the component. If I do not wrap the whole body - e.g. only a single child - then it is ok.

## Analysis

The element picker goes app the tree of DOM Elements starting with the mousover target until it find one that is a react component (has prop `__reactFiber$<something>` - whose value is the _raw-react-instance_ and underneath it `.return.stateNode` - whose value we call  _instance_) AND a fulcro component (there is the prop `instance.fulcro$isComponent = true`).

Given these two components:

```clojure
(ns ui (:require [com.fulcrologic.fulcro.react.error-boundaries :as eb]))
(defsc UboundedChild [_ _] {} (p :#UboundedChild "UboundedChild"))
(defsc ChildWithBoundary [_ _] {} (eb/error-boundary (p :#ChildWithBoundary "ChildWithBoundary")))
```

For **UboundedChild** it works and the picker's `over-handler` can see the following when it looks at the event's target:

* `(raw-react-instance target)`:
  * `.stateNode: p#unboundedChild`
  * `.return.stateNode`: ui$UnboundedChild
    * `.props: {fulcro$value: ...}`, `.fulcro$mounted`, `.fulcro$isComponent`
    * `.elementType: UnboundedChild`
* => target -> raw-react-instance -> .return.stateNode ->  fulcro$isComponent is true

But **ChildWithBoundary** is not detected as a Fulcro component:
* `(raw-react-instance target)` ; target =`<p id=ChildWithBoundary>...`
  * `.stateNode p#ChildWithBoundary` (consistent with the OK case)
  * `.return`
    * `.stateNode`: **nil** (in OK it was "ui$UnboundedChild")
    * `.type: eb/BodyContainer`
    * `.return`
      * `.type: eb/ErrorBoundary`
      * `.return.stateNode`: ui.UnboundedChild
        * `.props: {fulcro$value: ...}`, `.fulcro$mounted`, `.fulcro$isComponent`
* => target -> raw-react-instance -> .return.stateNode is nil but:
  * `.type = eb/BodyContainer`
  * `.return.return.stateNode.fulcro$isComponent` is true

### Conclusion

I.e. instead of checking just `raw-react-instance.return.stateNode`
we need to look also at `raw-react-instance.return[.type=eb/BodyContainer].return.return.stateNode`

holyjak:
Enhance the element picker to find also components that wrap their body in `eb/error-boundary` (only react 16+)

### DIFF
--- a/src/main/com/fulcrologic/fulcro/inspect/element_picker.cljc
+++ b/src/main/com/fulcrologic/fulcro/inspect/element_picker.cljc
@@ -6,7 +6,8 @@
                [goog.style :as gstyle]])
     [taoensso.timbre :as log]
     [com.fulcrologic.fulcro.components :as comp]
-    [com.fulcrologic.fulcro.inspect.inspect-client :as inspect]))
+    [com.fulcrologic.fulcro.inspect.inspect-client :as inspect]
+    [com.fulcrologic.fulcro.react.error-boundaries :as eb]))
 
 (def base-me-style #js {:position       "absolute"
                         :display        "block"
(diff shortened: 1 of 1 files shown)


holyjak:
**BEWARE** Likely due to a fulcro change between 1af65250...425567e2 (i.e. between in 3.5.8), Element Picker now fails because it cannot find `js/ReactDOM` herehttps://github.com/fulcrologic/fulcro/blob/86d963ff144a709b05b0d1128665c5b07b04d425/src/main/com/fulcrologic/fulcro/inspect/element_picker.cljc#L103

PR #502, opened 2022-01-15
Pass user-specified `post-mutation-params` in state-machine load

dvingo:
The parameters passed to `load-actor` for example were not passing
along user-supplied `post-mutation-params`.


awkay:
Good catch

awkay:
Released in 3.5.12

PR #506, opened 2022-02-07
Pass parameters from change-route! to will-enter

dvingo:
Hi Tony, I am working on a fulcro integration with reitit for routing and one of the features of reitit is the ability to coerce parameters to different data types from strings using declarative data on a route. I would like to use these already coerced values when implementing fulcro's `:will-enter` function.
 When reitit's on-url-change handler is fired you have access to the coerced parameters, which are passed to fulcro's `change-route!`, like: 
```clojure
(change-route! fulcro-app ["task" #uuid "017ed046-aa5b-8f6c-bfda-fa1b0ada980c" "edit"]
 {:task-id #uuid"017ed046-aa5b-8f6c-bfda-fa1b0ada980c"})
```
And a component would declare its `route-segment`: `[:task-id "edit"]` (and a parent route that handles the "task" prefix).

The changes in this PR will pass through the values provided to `change-route!` on to `:will-enter`. This way the parameters
don't need to be coerced again from strings inside of `will-enter`.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
+++ b/src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
@@ -381,6 +381,19 @@
                               :timeout! {::uism/handler identity}
                               :route!   {::uism/handler route-handler}}}}})
 
+(defn params-for-will-enter
+  "Returns a map that will be passed as the second parameter to a route target's :will-enter function.
+  Uses the parameter value passed to `change-route!` if it is present, otherwise using the value from the matching-prefix."
+  [change-route-params segment matching-prefix]
+  (reduce
+    (fn [p [k v]]
+      (if (keyword? k)
+        (let [value (if (contains? p k) (k p) v)]
+          (assoc p k value))
+        p))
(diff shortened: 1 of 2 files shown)


awkay:
Um...I don't really have the time to analyze this one for breakage. too busy right now. I'll have to sit on it for a while.

dvingo:
I have a satisfactory solution using a high-order helper to get this working, so I don't have strong need for this change if you'd like to close it.

awkay:
Yeah, I'm not sure this makes sense...you're saying if I have a route "/a/:task/b" and I route to "/a/999/b" but pass params map {:task 100} it will *really* act as-if the URL was "/a/100/b", which would be very misleading to the end user. It saves you on coercion, but it isn't a clean solution...it's rather a hack, so yeah, this isn't going in, but thanks for the submission!

dvingo:
Aha, that isn't something I was thinking of - more along the lines of `"999" -> 999 (number type)`,  but with this change what you demonstrated would be supported. So, I agree with your assessment - also don't have to deal with adding any breaking changes : )

PR #507, opened 2022-02-08
Add error check for form fields that are not in props

dvingo:
Was debugging a form component and realized I had a typo in the form-fields set, this will make the mistake an error so it can be caught earlier.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
@@ -115,7 +115,12 @@
                {:offending-component class})))
     (when (and (not has-fields?) queries-for-config?)
       (throw (ex-info (str "Attempt to add form configuration to " (rc/component-name class) ", but it does not declare any fields!")
-               {:offending-component class})))
+              {:offending-component class})))
+    (let [extra-fields (seq (set/difference all-fields fields subform-keys))]
+      (when (and has-fields? extra-fields)
+        (throw (ex-info (str "Attempt to add form configuration to " (rc/component-name class) ". It declares fields but not all of them are in the query!"
+                          "\n\nThe following fields are not in the query: \n\n" (pr-str extra-fields) "\n")
+                 {:offending-component class}))))
     [fields subforms subform-keys]))
 
 (>defn add-form-config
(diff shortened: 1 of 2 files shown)


src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
```diff
@@ -116,6 +116,9 @@
     (when (and (not has-fields?) queries-for-config?)
       (throw (ex-info (str "Attempt to add form configuration to " (rc/component-name class) ", but it does not declare any fields!")
                {:offending-component class})))
+    (when (and has-fields? (empty? fields))
```
awkay, line 119: what about subform-keys? They couldbe subforms

dvingo: thanks, I'll look into this.

dvingo: haven't looked into this yet, might have some time in the next few days though

dvingo:
Hi Tony, got a chance to update this, it is now ready for review.

src/test/com/fulcrologic/fulcro/algorithms/form_state_spec.cljc
```diff
@@ -148,6 +163,10 @@
   (component "error checking"
     (let [data-tree {:id 1 :name "A" :thing {:id 2 :x 42}}]
       (assertions
+        "throws an exception if any declared form fields are missing from the component query"
+        (fs/add-form-config FormMissingFields data-tree) =throws=> #"to .*FormMissingFields. It declares fields but not all of them are in the query!"
+        "does not throw an exception if all fields are present in the query when some are sub-forms"
```
awkay, line 168: In general I recommend using positive language in test assertions. For example, "Can successfully generate the config for a valid form with subforms".  Saying it doesn't do something opens the door to everything in the universe (conceptually). It doesn't paint my house, wash my car, re-render the screen, launch a rocket, etc...saying what it doesn't do is basically saying nothing at all.  

In the case of exceptions I guess it is a grey area, but I think it is clearer to say that a good case succeeds and a bad case fails.

no need to change anything, I just thought I would pass that on as food for thought.

dvingo: Heh, I had the same thought. If `add-form-config` is used with any subform that would verify this works. I think adding this under a positive test for subforms would be the way to go for something like this in the future.

Thanks for pointing this out!

PR #511, opened 2022-03-14
Fix targeting/replace-at for [] target

holyjak:
If the target property's value was a vector then replace-at did nothing,
instead of replacing the whole value.


awkay:
Released in 3.5.15

PR #512, opened 2022-04-06
In use-component, clear prior-props when current-ident changes

danieroux:
Allows fdn/possibly-stale? to become true.

Without this change, we have seen current-ident changing, prior-props staying the same, and the component therefore never updates.


awkay:
Good catch! Thanks!

awkay:
Released in 3.5.16

PR #514, opened 2022-04-12
Use (-> app config :load-mutation) in df/load-field! like df/load!

markusalbertgraf:
Hello Toni!
I tweaked `finish-load!` to trigger persisting data to IndexedDB.
So I provide a copy of `internal-load!`, which uses the tweaked `finish-load!` as `:load-mutation` to my app.
This works like a charm with `load!` but gets ignored by `load-field!`.
This patches `load-field!` to behave like `load!` in respect to honouring the `:load-mutation` in app/config.
This might be a breaking change if someone relies on `load-field!` ignoring `:load-mutation`.
Thank you for your work, fulcro and your training and education!


awkay:
I don't consider that breaking. It fixes a bug no one else has run into is my thinking. Thanks!

PR #516, opened 2022-05-16
Add :component-will-unmount-middleware option to fulcro-app

dvingo:
This option allows wrapping `componentWillUnmount` calls created by `defsc`
class components, which allows for performing arbitrary component cleanup
across an entire fulcro application without needing to create a `defsc` wrapper
macro.

I'm working on a rendering library integration with fulcro which requires performing cleanup for every component in the application.
Not having this middleware currently means I have to copy the defsc implementation from the components namespace (and copy many of its helper functions which are private), which would all go away with this helper.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/lookup.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/lookup.cljc
@@ -30,6 +30,7 @@
   - `:index-component!` - The algorithm that adds a component to indexes when it mounts.
   - `:drop-component!` - The algorithm that removes a component from indexes when it unmounts.
   - `:props-middleware` - Middleware that can modify `props` for all components.
+  - `:component-will-unmount-middleware` - Middleware that wraps `componentWillUnmount` methods of `defsc` components.
   - `:render-middleware` - Middlware that wraps all `render` methods of `defsc` components.
 
   Returns nil if the algorithm is currently undefined.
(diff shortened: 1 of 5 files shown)


awkay:
Hi Dan,

I'm excited to hear you are adding something to the ecosystem that you're excited about; however, this is going to be a bit of a hard sell, I'm afraid, for the following reasons:

* It ties to the idea you're using React or something that implements the legacy lifecycle of React
* It doesn't work with hooks components, because they have no such lifecycle method. Will hooks work with this rendering library? Why/why not? If not, it is an additional asymmetry in the library that is hard for users to understand.
* Your particular use-case is doable with a simple wrapper ns. I *could* be convinced to make the macro implementation functions public. That's a much easier sell. I'm still not sure how it might work with hooks, or if it is even intended to.
* It is a slippery slope of "Why not have a wrapper for EVERY lifecycle method"
* The unnamed "rendering library" is vaporware to me. Why should I be motivated to make a special-case exception for your "itch"? Is there something amazing you're bringing to the ecosystem?

awkay:
So, I've explored some ideas for derived data, and I'm glad you're looking at the problem. I agree that having a derived data "plug in" for Fulcro could be nice, but I also think you might be going about it in a less than optimal way: copying the ideas from Reframe are not what I would do...though I would be interested in seeing what it is you've actually done.

From my perspective one of the biggest problems with this model is that people are tempted to tie non-UI concerns to such a chain (e.g. I/O...doing loads and such in these event chains), which I think is a very bad and generally non-composable thing.

I think perhaps there are much more elegant ways of doing true (and pure) derived state in Fulcro. Perhaps you've already written a thing that you've got in production use, so you'd not be all the amenable to changing course, but if you are interested in hearing my thoughts on this particular area I'd love to schedule a chat with you.

dvingo:
Thanks Tony! I'm going to close this. _(for the github record - Tony and I had a chat - drop-index! can be used here)_

PR #517, opened 2022-05-30
Support named components in raw union queries

dvingo:
fulcro raw.components: Use a named component in union queries if one is present.


awkay:
Nice catch. Thanks.

PR #519, opened 2022-09-15
Accept list in addition to vectors for to-many join values

holyjak:
Problem
-------

If a resolver returns a lazy list inside the value, as in `{:friends (filter friend? people)}` then this will be transmitted as a list and finally considered an invalid join value and replaced with `{}`.

This repeatedly trips people. They see they are returning data from the backend, even see it in the brower's Network tab, but client DB ends up just with `{}` and they have no idea why the data disappeared. Noticing that they are actually returning a (lazy) list and that they should be returning a vector is not simple.

Solution space
--------------

1. We could log a warning. That would hopefully make it far easier to troubleshoot the problem, even if people keep running into it.
2. Ask Pathom or Fulcro's Transit integration to turn all lists into vectors automatically. Cons: it does not really seem that the solution belongs there since the problem is inside `merge`.
3. Accept `sequential?` instead of just `vector?`, thus making it work seamlessly (it gets later converted to a vector anyway). Pros: Simple change, user friendly. Cons: Possible unintended consequences???

Opted for 3. as Tony sees the risk as acceptable and it is the cleanest and most user-friendly solution.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/merge.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/merge.cljc
@@ -128,15 +128,15 @@
                   (assoc result jk ::not-found)
 
                   ; to-many join
-                  (and join? (vector? (get result jk)))
+                  (and join? (sequential? (get result jk)))
                   (assoc result jk (mapv (fn [item] (mark-missing-impl item (util/join-value element))) (get result jk)))
 
                   ; to-one join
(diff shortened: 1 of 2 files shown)


holyjak:
I did not run the tests, [Google Actions did](https://github.com/fulcrologic/fulcro/actions/runs/3072883609/jobs/4964657398) :)

I have added the test now, thank you for pointing that out!

PR #521, opened 2022-10-05
Improve Synchronous TX Processor behavior in CLJ

JJ-Atkinson:
This changes `submit-sync-tx!` to block if
 - the current invocation is "top level", subsequently clearing out the accumulated TX queue and returning (original behavior)
 - some other thread is clearing the TX queue, in which case it just spins and awaits completion of processing on the other thread

To compliment this, the depth tracking changed a little to include information about _which_ thread is currently running the queue, which is required to prevent deadlocks.

Verified in our current test suite to greatly improve reproducibility.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/tx_processing/synchronous_tx_processing.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/tx_processing/synchronous_tx_processing.cljc
@@ -83,19 +83,26 @@
 
 (declare run-queue! available-work?)
 
+(defn current-thread-id 
+  "Get the current thread id on the JVM. Returns 0 on JS."
+  []
+  #?(:clj (.getId (Thread/currentThread))
+     :cljs 0))
+
 #?(:clj
    (defmacro in-transaction [app-sym & body]
      `(let [id# (:com.fulcrologic.fulcro.application/id ~app-sym)]
-        (swap! apps-in-tx update id# (fnil inc 0))
+        (swap! apps-in-tx update id# conj (current-thread-id))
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/algorithms/tx_processing/synchronous_tx_processing.cljc
```diff
@@ -83,19 +83,26 @@
...
+  "Get the current thread id on the JVM. Returns 0 on JS."
+  []
+  #?(:clj (.getId (Thread/currentThread))
+     :cljs 0))
+
 #?(:clj
    (defmacro in-transaction [app-sym & body]
      `(let [id# (:com.fulcrologic.fulcro.application/id ~app-sym)]
-        (swap! apps-in-tx update id# (fnil inc 0))
+        (swap! apps-in-tx update id# conj (current-thread-id))
```
awkay, line 95: `(fnil conj #{})` I think. Or are you creating a stack? If the latter, then `(fnil conj [])`, which would obviate the need for `vec` conversion in the finally step

JJ-Atkinson: It needs to be a vec since it's a stack. I'll change to `(fnil conj [])` 👍

JJ-Atkinson: I just realized that `conj`/`pop` is the actual pair I wanted, so I've just changed to that instead.

PR #522, opened 2022-10-17
Support components with dynamic queries in add-form-state

dcostaras:
I'd like to use `add-form-config` with a component that utilises dynamic queries and as such need to pass app state to the get query call. This change, while maybe not the ideal API, (`:state-map` as an arg to `opts` which is not a pattern used elsewhere in Fulcro I don't think) is backwards compatible with the existing API.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
@@ -90,9 +90,9 @@
                  (map (fn [[k v]] {k (with-meta {} {:component v})}))
                  subforms)}))
 
-(defn- derive-form-info [class]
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/algorithms/form_state.cljc
```diff
@@ -140,7 +140,7 @@
    (add-form-config class entity {}))
```
awkay, line 140: Could use an update on the docstring of this...otherwise, good catch!

dcostaras:
@awkay Hi Tony, apologies, came here to update the docstring and saw it was done. Been grinding towards a deadline. Thanks for the merge.

PR #525, opened 2023-01-09
Remove extraneous clojure.pprint require

nivekuil:
noticed this added 100kb to the build output


awkay:
ooops. Thanks. Good catch.

awkay (approved):

PR #532, opened 2023-02-11
Add accept header because default client only accepts transit

### LINKED ISSUE #531: Potentially add "Accept: application/transit+json" header.
https://github.com/fulcrologic/fulcro/blob/18c27e5ddb694763e03f846bda8582d8eb749822/src/main/com/fulcrologic/fulcro/networking/http_remote.cljs#L142

The twin handler also assumes/only works with transit. This tripped me up because muuntaja saw the default "Accept: */*" header in the request and returned json. I think that "Accept: */*" can be argued to be a wrong header, because the my handler (the default fulcro one) only accepts transit.

I know fulcro is not tied to transit, but I think this is a reasonable default to add. Especially because this fails quietly when you use muuntaja, another popular clojure library which handles content negotiation

tommy-mor:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

fixes #531

### DIFF
--- a/src/main/com/fulcrologic/fulcro/networking/http_remote.cljs
+++ b/src/main/com/fulcrologic/fulcro/networking/http_remote.cljs
@@ -139,7 +139,7 @@
      (fn [{:keys [headers body] :as request}]
        (let [[body response-type] (desired-response-type request)
              body    (ct/write writer body)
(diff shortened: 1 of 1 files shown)


awkay (approved):

awkay:
Thanks!

PR #534, opened 2023-02-26
missing ^js tag causing hot-reloading to fail due to warning

yenda:


src/main/com/fulcrologic/fulcro/networking/http_remote.cljs
```diff
@@ -359,7 +359,7 @@
                                                      (request-middleware {:headers {} :body base-body :url url :method :post})
                                                      (catch :default e
                                                        (log/error e "Send aborted due to middleware failure. Did you accidentally put a lambda in params? See https://book.fulcrologic.com/#err-httpr-send-abort")
-                                                       (when-let [errant-data (some-> e (.-data) (.-obj))]
+                                                       (when-let [errant-data (some-> e (.-data) ^js (.-obj))]
```
awkay, line 362: Should probably annotate the `e` at catch, not in the middle of an expression like this

yenda: ok done

awkay (approved):

PR #535, opened 2023-03-20
fix: ensure removing source works with app-state path

yenda:
when the source-path isn't an ident and `remove-source?` is true, the expected behavior of targeting/process target is to remove the thing at the `source-path` from the app-state.
`dissoc` will not work with an app-state path.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/data_targeting.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/data_targeting.cljc
@@ -5,6 +5,7 @@
     [clojure.spec.alpha :as s]
     [com.fulcrologic.guardrails.core :as gw :refer [>defn => >def]]
     [edn-query-language.core :as eql]
+    [taoensso.encore :as enc]
     [taoensso.timbre :as log]))
 
 (>def ::target vector?)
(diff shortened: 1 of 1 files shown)


src/main/com/fulcrologic/fulcro/algorithms/data_targeting.cljc
```diff
@@ -153,4 +154,10 @@
                          (log/warn "Target processing found an unsupported case. See https://book.fulcrologic.com/#warn-target-unsuported-case")
                          state-map))))]
      (cond-> (process-target-impl state-map source-path target)
-       (and remove-source? (not (eql/ident? source-path))) (dissoc source-path)))))
+       (and remove-source?
+            (keyword? source-path))
+       (dissoc source-path)
+       (and remove-source?
```
awkay, line 160: When would the source path *ever* be a vector but not an ident? When you load, you either use a keyword, or you use an ident. You should never be issuing a load with some strange vector path. Can you explain where this is useful?

awkay (changes requested):

yenda:
The docstring itself mentions it:

"A keyword, ident, or app-state path"

In my case I ended up not using it because I wanted to merge not replace
but I was trying to use it to move a placeholder from one ident to another

src/main/com/fulcrologic/fulcro/algorithms/data_targeting.cljc
yenda: The docstring itself mentions it:

"A keyword, ident, or app-state path"

In my case I ended up not using it because I wanted to merge not replace
but I was trying to use it to move a placeholder from one ident to another

awkay: Hm. Clearly I forgot supporting that case. I'm not sure what I was thinking at the time, but since I documented it to work that way, I won't second-guess it. Thanks for the work.

awkay (approved):

PR #539, opened 2023-04-25
react18: assoc reactRoot to fulcro app

qleguennec:
I haven't seen any way to grab the react 18 root from a mounted fulcro app, so here it is.


awkay:
Did you try using `(app/app-root app)`?  It should work already

PR #544, opened 2023-08-02
Fix ident & link queries when target value is `false`

### LINKED ISSUE #543: Link/ident queries don't work with `false` values
When using a link or ident query, nothing is returned when the value at the queried-for ident is `false`. For example, in the example below, the `Child` component will receive an empty props map instead of the expected `{:root/pred? false}` and will therefore render nothing instead of `"false"`:

```Clojure
(defsc Child [this {:keys [root/pred?]}]
  {:query         [[:root/pred? '_]]
   :initial-state {}}
  (dom/div (str pred?)))

(def ui-child (comp/factory Child))

(defsc Root [this {:keys [root/pred? child]}]
  {:query         [:root/pred? {:child (comp/get-query Child)}]
   :initial-state (fn [_] {:root/pred? false
                           :child (comp/get-initial-state Child)})}
  (ui-child child))
```

In use cases where we only care about truthy values for the prop, this won't matter, but it becomes relevant when there's a meaningful distinction between `false` and `nil` or missing values.

joncurrier:
Resolves #543

### DIFF
--- a/src/main/com/fulcrologic/fulcro/algorithms/denormalize.cljc
+++ b/src/main/com/fulcrologic/fulcro/algorithms/denormalize.cljc
@@ -61,9 +61,10 @@
   (reduce
     (fn [n {:keys [key]}]
       (if (lookup-ref? key)
-        (if-let [x (follow-ref state-map key)]
-          (assoc! n (ref-key key) x)
-          n)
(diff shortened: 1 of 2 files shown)


awkay:
Excellent. Thanks!

PR #547, opened 2023-09-23
Mark render-to-str deprecated, show how to do it yourself

holyjak:
Use proper require of react-dom/server and invoke its renderToString, instead of accessing the global js/ReactDOMServer, which might not be available.

I had troubles making sure that `js/ReactDOMServer` and thus couldn't render to a string. With this change, it works fine.

**BEWARE**: I don't know whether the use of js/ReactDOMServer is a left-over from old times, or whether it is intentional, e.g. to make sure that we do not require `react-dom/server` if not needed. So I don't know whether this change is desirable or not.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/dom.cljs
+++ b/src/main/com/fulcrologic/fulcro/dom.cljs
@@ -71,8 +71,14 @@
   [component el]
   (react.dom/render component el))
 
-(defn render-to-str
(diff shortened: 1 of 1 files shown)


awkay:
I think I had some notion that I wanted to make this optional...to avoid code bloat. shadow will make that an extern, and will include it in every build (even if you don't use it. no dead code elimination on js libs). Doing the way I did it means you won't get react dom server in your advanced builds unless you require it elsewhere.

so, while what you have in cleaner, I'd rather see a deprecated tag on the function with a recommendation that you just use it yourself.

holyjak:
Done!

PR #548, opened 2023-11-03
fix: render with force-root? option not working in release builds

yenda:


awkay:
So the intention of this option was really for dev-time hot code reload. If you're calling this as part of production code, then you're using it wrong. So, I'm not sure I do want to make this work in non-dev builds.

awkay:
Fixed and released, but this wasn't the correct patch.

PR #550, opened 2024-01-20
Fix #549 Propagate parent to hooks components

### LINKED ISSUE #549: Hook-based defsc should also have their parent set
Inside a defsc with `:use-hooks? true`, the `(comp/get-parent this)` call returns `nil` - contrary to non-hook componets. It should work in both kinds of components.

(Note: I have tested this only on router targets, but assume it behaves the same everywhere.

holyjak:
Issue: The render body of hooks-based components gets its props from the use-fulcro hook, which failed to propagate `fulcro$parent` in the raw JS props. comp/get-parent relies on it, and thus stopped working there.

Fix: Add it to the manually tunnelled raw JS props in use-fulcro.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/components.cljc
@@ -495,10 +495,12 @@
            current-state           (aget tunnelled-props-state 0 "fulcro$value")
            props                   (isoget js-props :fulcro$value)
            children                (isoget js-props :children)
+           parent                  (isoget js-props :fulcro$parent)
            current-props           (newer-props props current-state)
            current-ident           (when ident (ident faux-class current-props))
            shared-props            (when app (shared app))]
        (doto (gobj/get faux-component "props")
+         (gobj/set "fulcro$parent" parent)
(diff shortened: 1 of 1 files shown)


holyjak:
Seems to work, but in combination with fulcro-troubleshooting in rad-demo it causes

> error rendering com.example.ui/MainRouter: Maximum call stack size exceeded

and I want to figure out why first. Perhaps I somewhere set component's parent to be itself?

holyjak:
FYI: The `RangeError: Maximum call stack size exceeded` I got was caused by the fact that I had `(pr-str (comp/get-parent this))` in the (always rendewring) body of the MainRouter, and the data was too large. When I sent print-length to 10 and print-level to 50, it did work, but printed some 1.6MB data. With level 100 the browser just crashed 😅. I would like to understand why it is so large, even if we know it contains the whole fulcro app multiple times (in the parent, and its parent....). But printing the whole `this` sounds anyway as somwething you should never do.

holyjak:
Hm, the problem I found seems unrelated to this change.

awkay (approved):

PR #551, opened 2024-01-23
Make sure fulcro app prints safely

holyjak:
Issue: Printing a fulcro-app could blow the stack because an atom could contain itself. And even if it didn't, the app is huge and you rarely really want to print it.

Fix: Take inspiration from Component and replace the map with a record, for which we implement safe print-method. As a bonus, this also simplifies detecting whether something is a Fulcro app.

Here we see how it prints an different cases:

Screenshot 2024-01-23 at 22 17 33


awkay:
Did you try this out against anything of size? Like a RAD demo app?

holyjak:
Yes, I tried it in rad-demo.

awkay (approved):

PR #555, opened 2024-10-27
Improve error message when a map passed by mistake as a child to fragment

holyjak:
... or to ErrorBoundary (which uses comp/fragment).

Without this change, we assume the map is props, and end up getting a _warning_ like "Warning: Invalid prop `some-prop...` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props."

With this change, we get a proper _error_ about trying to render the wrong type, and in which component.

### DIFF
--- a/src/main/com/fulcrologic/fulcro/components.cljc
+++ b/src/main/com/fulcrologic/fulcro/components.cljc
@@ -1020,7 +1020,9 @@
                           [{} args])]
        (vec children))
      :cljs
-     (let [[props children] (if (map? (first args))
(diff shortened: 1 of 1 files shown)


holyjak:
Testing: A simple app where Root has an arbitrary map as its body.

Of course a disadvantage of this change is that if fragment adds support for more properties in the future then we'd need to update the code. I am not sure whether we want to accept the risk or not.

awkay (commented):
React MAY add additional attributes in the future. I don't think this is what we should do.

Issue #556, opened 2024-12-07
df/load! makes query invalid for components with a union query


scizo:
First off, thanks for Fulcro. We have just been getting started with it, but it is clear the that a lot of care and thought have gone into it!

When using `df/load!` with a component that has a union query, the query gets altered into an invalid query before being sent to server. Here is a diff for the data fetch specs with a failing test to demonstrate.

```diff
diff --git a/src/test/com/fulcrologic/fulcro/data_fetch_spec.cljc b/src/test/com/fulcrologic/fulcro/data_fetch_spec.cljc
index cf188b31..acea2b73 100644
--- a/src/test/com/fulcrologic/fulcro/data_fetch_spec.cljc
+++ b/src/test/com/fulcrologic/fulcro/data_fetch_spec.cljc
@@ -30,12 +30,22 @@
    :ident         [:parent/by-id :x]
    :query         [:x :z {:child (comp/get-query InitTestChild)}]})

+(defsc UnionComponent [this props]
+  {:initial-state (fn [params] {:a/id 1 :a/name "Alice"})
+   :ident (fn [] (cond
+                   (:a/id props) [:a/id (:a/id props)]
+                   (:b/id props) [:b/id (:b/id props)]
+                   :else nil))
+   :query (fn [] {:a/id [:a/id :a/name]
+                  :b/id [:b/id :b/another-prop]})})
+
 (def app (app/fulcro-app))

 (specification "Load parameters"
   (let [
         query-with-params       (:query (df/load-params* app :prop Person {:params {:n 1}}))
-        ident-query-with-params (:query (df/load-params* app [:person/by-id 1] Person {:params {:n 1}}))]
+        ident-query-with-params (:query (df/load-params* app [:person/by-id 1] Person {:params {:n 1}}))
+        union-query (:query (df/load-params* app :prop UnionComponent {:without #{}}))]
     (assertions
       "Accepts nil for subquery and params"
       (:query (df/load-params* app [:person/by-id 1] nil {})) => [[:person/by-id 1]]
@@ -48,7 +58,9 @@
       (:target (df/load-params* app :prop Person {:target [:a :b]})) => [:a :b]
       "Constructs a JOIN query (with params on join and prop)"
       query-with-params => `[({:prop ~(comp/get-query Person)} {:n 1})]
-      ident-query-with-params => `[({[:person/by-id 1] ~(comp/get-query Person)} {:n 1})]))
[...]

PR #557, opened 2024-12-09
df/load! properly handles components with a union query

scizo:
df/load! was altering component queries and making them invalid when
they were a union query at the top level component. More info can be
found in #556.

This commit wraps the component query in a temporary top level root
query before eliding attributes provided by the without option, which is
what was causing the problem.


awkay:
Thanks

scizo:
Thank you!

PR #558, opened 2025-01-09
'|' is not permit for filename in Windows

free-kingdom:
Hi, I found that using the '|' character in file-upload caused issues on Windows, as it's not allowed in filenames. I replaced it with %, and my code works (I test it in both Debian and Windows). Is the char just used internally and the change will not break other piece of code?


awkay:
Thanks. Didn't realize that. Simple enough fix!

Issue #559, opened 2025-01-24
Mismatched keys for alternative routes inside of router state when using `dr/dynamic-router`


scizo:
`defrouter` generates a query and initial state that gets state for alternative targets by creating a key like `:alt0`, `:alt1`, etc. `dynamic-router` also generates these same keys for the initial-state, but it generates a query using namespaced keys, like `:alt/0`, `:alt/1`, etc. This causes a warning to be logged by `rc/check-component-registry!` because the initial-state does not match the query, so it thinks the initial-state wasn't generated. I am assuming that the namespaced keys were unintentional.

awkay:
Fixed. Thanks!

PR #560, opened 2025-01-24
Use same key convention for alternate route keys in `dr/dynamic-router`

scizo:
See issue #559 for more details on the bug. This commit opts for
creating a utility function to create the keys, not because it isn't
easily inlined, as was done before, but I thought it would codify and
reduce the chances of a similar mismatch if this is needed again.


awkay:
Great catch...I never create routers with that helper function. This explains my confusion...the macro is fine, It's the dyn one that is hosed.

scizo:
Thanks @awkay! I appreciate all that you do and for sharing all of this wonderful code and time.

PR #561, opened 2025-03-14
Bugfix: Dissoc ident after using it

evaogbe:
In the old implementation, the ident of `defnc` was always nil. This is because the ident was dissoced before it was being used. This change moves the dissoc to after the ident is used, so that it can be later assoced.


awkay:
in 3.8.3

PR #562, opened 2025-03-21
Allow `defrouter` to `:use-hooks?`

danieroux:


src/main/com/fulcrologic/fulcro/raw/components.cljc
```diff
@@ -358,7 +358,10 @@
                        (-> class-or-factory meta :qualifier)
                        nil)
            queryid   (if (component-instance? class-or-factory)
-                       (get-query-id class-or-factory)
+                       ; hooked use-fulcro components are component instances, without comp/factory's :fulcro$queryid
+                       (or
```
danieroux, line 362: @awkay I am unclear if this is a workaround or a bugfix

awkay: Good question. Dynamic queries were written before raw and hooks. I don't see how your change hurts anything. I mean, I guess we could allow a query id to be set when using the hook? That's kind of a double-level of dynamism I'm not sure is great, so let's go with this...

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
```diff
@@ -934,15 +935,21 @@
                                                                               :current-state        ~'current-state}]
                                                        ~@body)))))
            options                (merge
-                                    `{:componentDidMount (fn [this#] (validate-route-targets this#))}
+                                    `{:componentDidMount (fn [this#] (validate-route-targets this#))
```
awkay, line 938: so if use hooks is true, you just leave this hook here, and are happy because it is a no-op? It's a little dangerous in that what if we add error checking to defsc around hooks to warn you when you do something dumb? Maybe a cond-> or something?

src/workspaces/com/fulcrologic/fulcro/cards/nested_dynamic_routing_tree_cards.cljs
```diff
@@ -108,16 +109,28 @@
 
 (def ui-a (comp/factory A {:keyfn :id}))
 
-(defrouter RootRouter [this props]
-  {:router-targets [A B]})
+(defrouter RootRouter [this {:keys [route-factory route-props]}]
```
awkay, line 112: Might consider making a new card so the old test card is still available for quick checking the non-hooks

src/main/com/fulcrologic/fulcro/routing/dynamic_routing.cljc
danieroux: Something like [this?](https://github.com/fulcrologic/fulcro/compare/8c01463a88fb867bfe18229629d1cd3c6f3c860e..c8dbb7851f5321bce91c15f9f00bf3b96b75dc92#diff-dbc631589940e7571a88fa512a517d49cc6e731bcaaa85c04b95de1c4d7df07bR954)

awkay: yes

danieroux:
Added cards testing router-rendering:

- Router that is a React class component
- Router created by `dr/dynamic-router`, class component
- Router that is a React hooks component
- Router created by `dr/dynamic-router`, hooks component

PR #563, opened 2025-03-22
Fix CI: GitHub `actions/cache@v2` -> ` actions/cache@v4` and `clj-doc` exclusion

danieroux:
- Deprecation notice for `actions/cache@v2`: https://github.com/actions/toolkit/discussions/1890
- Add `make-clj-doc` and fix `analyze-local`


awkay:
Thanks!

PR #565, opened 2025-04-24
Add `args` parameter to `use-imperative-handle` for dependency management

RKDGT:
Cause:
The current implementation of use-imperative-handle is missing the args parameter, which is similar in behavior to the dependency array in use-effect. Without this parameter, updates to dependencies are not tracked, potentially causing stale references and unexpected behavior.

Fix:
Added an args parameter to the use-imperative-handle implementation, enabling dependency tracking and ensuring the function updates correctly when dependencies change. This aligns its behavior with other React hooks like use-effect.


awkay:
Thanks

PR #569, opened 2025-08-07
fix wrapped select and option

free-kingdom:
forgot children in select and option...


awkay:
Thanks. THis is why it's an RC :D

